PR #6, opened 2021-03-04
IJ 29 devise views

IeuanSkinner:
Updated all of the devise views to use a centered bootstrap form template that includes floating labels for the input fields and the Include UK colours for the buttons and links.

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -1,13 +1,37 @@
 @import "variables";
 
 .ij-form {
+  display: flex;
+  min-height: 100vh;
+  align-items: center;
+  justify-content: center;
   text-align: center;
-  margin: auto;
+  background-color: $brand-secondary;
+
+  @media screen and (min-width: $screen-md) {
+    background-color: $bg-main;
+  }
 
   form {
     background-color: $brand-secondary;
     border-radius: 5px;
     padding: 15px 50px;
+
+    h2 {
+      color: white;
+    }
+
+    .col-md {
+      margin-top: 0;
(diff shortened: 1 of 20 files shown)


phil-6 (approved):
Some of the colours clash a little but this is an easy fix for later. Great work.

It looks as though there are some odd bugs with the unlocks controller but I think this can be resolved further down the line.

IeuanSkinner:
> Some of the colours clash a little but this is an easy fix for later. Great work.
> 
> It looks as though there are some odd bugs with the unlocks controller but I think this can be resolved further down the line.

I'm only using their brand colours so not sure what clash could be occurring unless you're saying their brand colours clash which I'd disagree with.

IeuanSkinner:
> Some of the colours clash a little but this is an easy fix for later. Great work.
> 
> It looks as though there are some odd bugs with the unlocks controller but I think this can be resolved further down the line.

For the unlocks controller have you dropped and created the DB with the latest changes? I also saw error messages from this before I'd merged the changes you'd put on master to enable the devise locking mechanism. 

With the changes those messages went aways and now on sign up you see 'uninitialized constant PagesController...' which is basically the result of some routing configurations that should get fixed via IJ-31

phil-6:
> > Some of the colours clash a little but this is an easy fix for later. Great work.
> > It looks as though there are some odd bugs with the unlocks controller but I think this can be resolved further down the line.
> 
> I'm only using their brand colours so not sure what clash could be occurring unless you're saying their brand colours clash which I'd disagree with.

The link blue and the alert red don't work well on such a dark background.

PR #7, opened 2021-03-04
IJ-31

IeuanSkinner:
Added in dashboard that the service user is redirected too upon successful authentication. I've also updated the user model and application controller so that the additional devise fields are also being set on the underlying record.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,4 +11,14 @@ class User < ApplicationRecord
   has_many :wba_team_members, foreign_key: true
 
   has_many :crisis_events, foreign_key: true
+
+  private
+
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -11,4 +11,14 @@ class User < ApplicationRecord
...
 
   has_many :crisis_events, foreign_key: true
+
+  private
+
+  # validations
+  validates_presence_of :first_name,
+                        :last_name,
+                        :email,
+                        :terms
```
phil-6, line 21: I think we want to require phone number as well.

phil-6 (approved):

PR #10, opened 2021-03-08
Feature/IJ-10 create seed

bmhtech07:
Created seed with all models with each of us as TeamMembers.

Note, priority on WbaTeamMemberScore is rand(1..3) - not sure what we're doing here.

Removed the foreign_key from CrisisEvent belongs_to (left original line commented out for reference). 

Added foreign_key to crisis_event migration to allow for the above relationship. 

Changed column 'type' to category as 'type' is reserved for devise (I believe).


app/models/crisis_event.rb
```diff
@@ -1,6 +1,7 @@
 class CrisisEvent < ApplicationRecord
   belongs_to :user
   belongs_to :crisis_type
-  belongs_to :closed_by, class_name: 'TeamMember', foreign_key: 'closed_by'
-  has_many :crisis_notes, foreign_key: true
+  belongs_to :closed_by, class_name: 'TeamMember', optional: true
+  # belongs_to :closed_by, class_name: 'TeamMember', foreign_key: 'closed_by' Original line
```
phil-6, line 5: This comment can be removed

phil-6 (approved):
Looks good to me

AlexW555 (approved):

db/seeds.rb
```diff
@@ -5,3 +5,324 @@
...
+  # id: 9
+  WellbeingMetric.create!(
+    team_member_id: 1,
+    name: 'Relationships',
+    category: 'Personal'
+  )
+  # id: 10
+  WellbeingMetric.create!(
+    team_member_id: 1,
+    name: 'Sense of Community ',
```
IeuanSkinner, line 150: Un-needed space at the end of name.

IeuanSkinner (changes requested):
Using a loop to generate both the WBA Self record and the scores linked to that record is smart and works a treat but then at the end of the seed you're using 2 disconnected loops to generate the WBA TM records and the scores linked to those records. As these loops are now disconnected you end up with the WBA TM Score table looking quite weird as the WBA TM ID column isn't increasing linearly as you'd expect, instead the values jump all over the place and you can't group them at a glance.

Programatically this probably won't affect anything but it makes reading the DB difficult and the fix just requires using the same logic as was done for WBA selves to be done for WBA team members.

IeuanSkinner (approved):
LGTM 👍

PR #12, opened 2021-03-08
Feature/ij 64 admin team management

IeuanSkinner:
When a new team member signs up they now require approval from an existing admin before they can actually login to the system. I've also added in the ability for an admin user to toggle the admin status of other team members.

### DIFF
--- a/app/controllers/team_members/dashboard_controller.rb
+++ b/app/controllers/team_members/dashboard_controller.rb
@@ -1,8 +1,8 @@
 module TeamMembers
   # app/controllers/team_members/dashboard_controller.rb
   class DashboardController < TeamMemberController
-    before_action :require_admin, only: %i[set_team_member, approve_team_member]
-    before_action :set_team_member, only: :approve_team_member
+    before_action :require_admin, only: %i[set_team_member, approve_team_member, toggle_admin]
+    before_action :set_team_member, only: %i[approve_team_member, toggle_admin]
 
     def main
       render template: 'team_members/dashboard/main'
@@ -12,9 +12,24 @@ module TeamMembers
     def approve_team_member
       respond_to do |format|
         format.html { redirect_to authenticated_team_member_root_path,
-                                  alert: @team_member.update({approved: true}) ?
-                                           "#{@team_member.first_name} has been approved!" :
-                                           "#{@team_member.first_name} could not be approved!" }
+                                  alert: @team_member.update(approved: true) ?
+                                           "#{@team_member.first_name} has been approved" :
+                                           "#{@team_member.first_name} could not be approved" }
+      end
+    end
+
+    # PUT /team_members/:team_member_id/admin
+    def toggle_admin
+      respond_to do |format|
+        if @team_member.update(admin: !@team_member.admin?)
+          format.html { redirect_to authenticated_team_member_root_path,
+                                    alert: @team_member.admin? ?
+                                             "#{@team_member.first_name} is now an admin" :
+                                             "#{@team_member.first_name} is no longer an admin" }
+        else
+          format.html { redirect_to authenticated_team_member_root_path,
+                                    warning: "#{@team_member.first_name} admin status could not be changed" }
+        end
       end
     end
 
--- a/app/controllers/team_members/dashboard_controller.rb
+++ b/app/controllers/team_members/dashboard_controller.rb
@@ -1,43 +1,25 @@
 module TeamMembers
   # app/controllers/team_members/dashboard_controller.rb
-  class DashboardController < TeamMemberController
-    before_action :require_admin, only: [:set_team_member, :approve_team_member, :toggle_admin]
-    before_action :set_team_member, only: [:approve_team_member, :toggle_admin]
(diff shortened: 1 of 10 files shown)


app/controllers/team_members/dashboard_controller.rb
```diff
@@ -1,8 +1,43 @@
...
+    before_action :set_team_member, only: [:approve_team_member, :toggle_admin]
+
     def main
       render template: 'team_members/dashboard/main'
     end
+
+    # PUT /team_members/:team_member_id/approve
+    def approve_team_member
+      respond_to do |format|
+        debugger
```
phil-6, line 14: this should probably not still be here

app/controllers/team_members/dashboard_controller.rb
```diff
@@ -1,8 +1,43 @@
...
+        format.html { redirect_to authenticated_team_member_root_path,
+                                  alert: @team_member.update(approved: true) ?
+                                           "#{@team_member.first_name} has been approved" :
+                                           "#{@team_member.first_name} could not be approved" }
```
phil-6, line 18: this is cool, but feels weird to put the update method inside the alert.

app/controllers/team_members/dashboard_controller.rb
```diff
@@ -1,8 +1,43 @@
...
+      end
+    end
+
+    # PUT /team_members/:team_member_id/admin
+    def toggle_admin
+      respond_to do |format|
+        if @team_member.update(admin: !@team_member.admin?)
```
phil-6, line 25: would this be easier to read with Unless instead of if?

app/controllers/team_members/dashboard_controller.rb
```diff
@@ -1,8 +1,43 @@
...
+          format.html { redirect_to authenticated_team_member_root_path,
+                                    alert: @team_member.admin? ?
```
phil-6, line 27: it feels nice to me to just have a check here instead of the actual `.update`

app/views/team_members/dashboard/main.html.erb
```diff
@@ -45,10 +45,42 @@
     </div>
   </section>
   <% if current_team_member.admin? %>
+    <% if TeamMember.all.where(approved: false).count > 0 %>
+      <section class="row" id="awaiting_approval">
+        <div class="col">
+          <h3>Awaiting Approval</h3>
```
phil-6, line 51: Maybe make this clearer that this is Team Members not Service Users?

app/views/team_members/dashboard/main.html.erb
```diff
@@ -45,10 +45,42 @@
     </div>
   </section>
   <% if current_team_member.admin? %>
+    <% if TeamMember.all.where(approved: false).count > 0 %>
```
phil-6, line 48: _Ideally_ we want to avoid making db calls from the view. I would add a private method which gets all the unapproved team members and call it in the the main method in the dashboard controller before the render call or as a before action.

app/views/team_members/dashboard/main.html.erb
```diff
@@ -57,16 +89,29 @@
             <th scope="col">Last Sign In</th>
           </thead>
           <tbody>
-            <% TeamMember.all.each do |team_member| %>
+            <% TeamMember.all.where(approved: true).order(:id).each do |team_member| %>
```
phil-6, line 92: As earlier comment, Model calls should be in the controller ideally

phil-6 (changes requested):
Looks good overall. Model actions should be moved to the controller instead of the view. Probably should have caught that on the last PR. 
Thanks Ieuan!

app/controllers/team_members/dashboard_controller.rb
IeuanSkinner: Personally I much prefer if, it just seems more intuitive to me to say "if update is successful do foo otherwise do bar" rather than "unless update is successful do bar otherwise do foo". I know they're logically identical but my thought process is much more used to the former.

app/views/team_members/dashboard/main.html.erb
```diff
@@ -44,39 +44,82 @@
...
+                <% if current_team_member.admin? and current_team_member.id != team_member.id %>
+                  <%= form_with url: toggle_admin_path(team_member.id), method: :put do |form| %>
+                    <%= form.hidden_field :team_member_id, value: team_member.id %>
+                    <%= button_to '', { action: 'submit', class: 'btn w-100'} do %>
+                      <% if team_member.admin %>
+                        <i class="fas fa-check-circle"></i>
+                      <% else %>
+                        <i class="fas fa-times-circle"></i>
+                      <% end %>
+                    <% end %>
```
phil-6, line 106: Do we want to add a confirmation dialogue to avoid accidental promotion?

phil-6 (approved):
Looks good, only comment is a future design thought.

app/views/team_members/dashboard/main.html.erb
IeuanSkinner: Done

PR #13, opened 2021-03-09
Feature/ij 26

AlexW555:
Universal footer across application with social links & Copyright
photo_2021-03-09_11-20-40

### DIFF
--- a/app/assets/stylesheets/pages.scss
+++ b/app/assets/stylesheets/pages.scss
@@ -23,3 +23,14 @@
   }
 
 }
+
+.footers{
+  .title {
+    color: white;
+    text-decoration: none;
+  }
+  a:hover{
+    text-decoration: underline;
+    color: grey;
+  }
+}
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -52,17 +52,16 @@
     <%= yield %>
 
   <footer class="text-center text-white bg-brand";>
-    <div class="container pt-4">
-      <a class="text-white" href="https://legaltech.wales/"><h3> Include uk</h3></a>
+    <div class="footers container pt-4">
+      <a class="title" href="https://include-uk.com"><h3> Include uk</h3></a>
       x
-      <a class="text-white" href="https://legaltech.wales/"><h3>Legal Innovation Lab Wales</h3></a>
+      <a class="title" href="https://legaltech.wales/"><h3>Legal Innovation Lab Wales</h3></a>
       <section class="mb-4">
-        <a class="btn btn-link btn-floating btn-lg text-dark m-1 bg-white" href="#!" role="button"
+        <a class="btn btn-link btn-floating btn-lg text-dark m-1 bg-white" href="https://www.facebook.com/includehub" role="button"
            data-mdb-ripple-color="dark"><i class="fab fa-facebook-f"></i></a>
-        <a class="btn btn-link btn-floating btn-lg text-dark m-1 bg-white" href="#!" role="button"
+        <a class="btn btn-link btn-floating btn-lg text-dark m-1 bg-white" href="https://include-uk.com" role="button"
           data-mdb-ripple-color="dark"><i class="fab fa-twitter"></i></a>
-
-        <a class="btn btn-link btn-floating btn-lg text-dark m-1 bg-white" href="#!" role="button"
+        <a class="btn btn-link btn-floating btn-lg text-dark m-1 bg-white" href="https://include-uk.com" role="button"
           data-mdb-ripple-color="dark"><i class="fab fa-instagram"></i></a>
       </section>
     </div>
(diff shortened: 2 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -51,5 +51,24 @@
 
     <%= yield %>
 
+  <footer class="text-center text-white bg-brand";>
```
phil-6, line 54: Stray semicolon

app/views/layouts/application.html.erb
```diff
@@ -51,5 +51,24 @@
...
+    <div class="footers container pt-4">
+      <a class="title" href="https://include-uk.com"><h3> Include uk</h3></a>
+      x
+      <a class="title" href="https://legaltech.wales/"><h3>Legal Innovation Lab Wales</h3></a>
+      <section class="mb-4">
+        <a class="btn btn-link btn-floating btn-lg text-dark m-1 bg-white" href="https://www.facebook.com/includehub" role="button"
+           data-mdb-ripple-color="dark"><i class="fab fa-facebook-f"></i></a>
+        <a class="btn btn-link btn-floating btn-lg text-dark m-1 bg-white" href="https://include-uk.com" role="button"
```
phil-6, line 62: Can we add the links to their twitter and insta? Instead of just their web?

phil-6 (commented):
Looks good Overall!

phil-6:
> The white on yellow is really difficult to look at, can we change the text and icons to use a dark colour if they're going to be sitting on a light colour?

I agree but this is the way its styled / branded on their home page.

> The 'x' between include uk and lilw seems out of place, I get that it's trying to infer the product is a collaboration between us and them but it doesn't do a good job of it.

I like it, I think it does need more styling. Maybe use at font-awesome Icon?

> This footer should function like the header and not appear on the devise views, currently it gets pushed off the screen as the view is min-height 100vh anyway, it too can be wrapped and not appear in the event of the user not being authenticated.

Instead of not appearing unless authenticated could we set it so it only doesn't appear on devise views. 
There is a way that you can do an IF based on the controller. `<%= if devise_controller? %>`

IeuanSkinner:
Also I want to highlight that the aversion to light-on-light / dark-on-dark layering isn't me trying to inject preference it's actually an anti-pattern with regards to accessibility as users with eyesight considerations (whether it be poor eyesight or certain colour blindness) will struggle to discern things when colours are layered in this way. I know it's something that Include UK do on their site but they really shouldn't.

IeuanSkinner:
> > The white on yellow is really difficult to look at, can we change the text and icons to use a dark colour if they're going to be sitting on a light colour?
> 
> I agree but this is the way its styled / branded on their home page.

Hadn't seen this before posting my above comment, I know it's how they've done things on their site but they're wrong to have done it there.

phil-6:
Yeah, your absolutely right. We can show them how it should be done! 
@AlexW555 Can you set them to $grey or $black for now please. I've created IJ-67 to capture any front end improvements we want to make before release which aren't urgent right now.

app/views/layouts/application.html.erb
AlexW555: I did plan to, but couldn't initially find their Twitter or Instagram when I searched for it

app/assets/stylesheets/pages.scss
```diff
@@ -23,3 +23,14 @@
...
 }
+
+.footers{
+  .title {
+    color: white;
+    text-decoration: none;
+  }
+  a:hover{
+    text-decoration: underline;
+    color: grey;
```
IeuanSkinner, line 34: We have a $grey variable setup that should be used here

app/assets/stylesheets/pages.scss
```diff
@@ -23,3 +23,14 @@
   }
 
 }
+
+.footers{
+  .title {
+    color: white;
```
IeuanSkinner, line 29: We have a $white variable setup that should be used here

IeuanSkinner: Though they should now be using a dark colour: https://github.com/Legal-Innovation-Lab-Wales/include-journey/pull/13#issuecomment-793804038

IeuanSkinner (approved):
Had a play with the colour scheme too and changed the links to use the plum black we use for the site background atop the include yellow. The hover colour is now the secondary candy pink which matches what the buttons do in the top section of the dashboard.

I'm happy with this as the final scheme 👍 
Screenshot 2021-03-11 at 15 01 42

phil-6:
I think I prefer the social buttons being white, although you lose the contrast with the background, the contrast is maintained with the icon.
Other than that, it looks good to me. 

On a more general colour scheme note, I'm still not a massive fan of the grey background for the main content. I feel like its quite bleak. But thats something for another time.

phil-6 (approved):

PR #14, opened 2021-03-09
Feature/ij 42 team member index

IeuanSkinner:
Admin team members can now drill down into an individual team member to view their details, sign in info and a list of WBA that team member has created as well as the logs for any user created WBA that the team member has viewed.

One enhancement we'll likely want in future is the ability to filter both of these tables by user and date, I could do that as a part of this ticket but for MVP just displaying all the info and using CTRL+F should suffice.

Screenshot 2021-03-09 at 11 46 02

### DIFF
--- a/app/views/team_members/team_members/main.html.erb
+++ b/app/views/team_members/team_members/main.html.erb
@@ -1,7 +1,115 @@
-<main class="container dashboard">
-  <section class="row">
+<main class="container dashboard team_member">
+  <section class="row" id="details">
     <div class="col">
-      <h3>This page is for <%= @team_member.email %></h3>
+      <h2 class="mb-3">Team Member Details</h2>
+      <div class="container">
+        <div class="row">
+          <div class="col">
+            <form>
+              <div class="row g-2">
+                <div class="col-md">
+                  <div class="form-floating">
+                    <input type="text-field" name="team_member_first_name" class="form-control" value="<%= @team_member.first_name %>" disabled>
+                    <label for="team_member_first_name">First Name</label>
+                  </div>
+                </div>
+
+                <div class="col-md">
+                  <div class="form-floating">
+                    <input type="text-field" name="team_member_last_name" class="form-control" value="<%= @team_member.last_name %>" disabled>
+                    <label for="team_member_last_name">Last Name</label>
+                  </div>
+                </div>
+              </div>
+
+              <div class="row g-2">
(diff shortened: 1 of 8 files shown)


app/views/team_members/team_members/main.html.erb
```diff
@@ -0,0 +1,106 @@
...
+        <table class="table table-dark table-striped">
+          <thead>
+            <th scope="col">User</th>
+            <th scope="col">Created At</th>
+            <th scope="col">Last Update</th>
+          </thead>
+          <tbody>
+            <% @team_member.wba_team_members.order(:created_at).each do |wba_team_member| %>
+              <tr>
+                <td><%= wba_team_member.user.first_name %> <%= wba_team_member.user.last_name %></td>
```
phil-6, line 73: Please can we add this as a link that will take to the team member user view. The page isn't built yet but it would be good to have the link there, ready.

app/views/team_members/team_members/main.html.erb
```diff
@@ -0,0 +1,106 @@
...
+        <h3>wellbeing assessment view logs</h3>
+        <table class="table table-dark table-striped">
+          <thead>
+          <th scope="col">User</th>
+          <th scope="col">Created At</th>
+          <th scope="col">Last Update</th>
+          </thead>
+          <tbody>
+          <% @team_member.wba_self_view_logs.order(:created_at).each do |wba_self_view_log| %>
+            <tr>
```
phil-6, line 95: This looks like its an n+1 query. 
In general these occur when you're looping over a model relationship. 
Its quite a common thing in Rails that can have a massive performance impact when tables get bigger. Worth a Google, but I believe we should do two things, `include` the data we need in the variable assignment and add pagination to this section so we don't return everything at once.

phil-6 (changes requested):
Overall looks good. Few small changes and a few rubocop recommendations. I would definitely consider do the "open in ruby mine first" trick I mentioned earlier as rubocops hints are genuinely useful and will make everything more maintainable in the long run.

app/views/team_members/team_members/main.html.erb
phil-6: It lost my comment when I clicked "Unresolve" sorry!
Maybe have the name link to the user and the table row link to the assessment, similar to the styling on the make admin butotn?

phil-6 (approved):

PR #15, opened 2021-03-09
Feature/ij 66 wellbeing chart

IeuanSkinner:
I've integrated chart.js along with chartkick into the platform and I've setup an initial design idea I had for the wellbeing chart to use a polar area chart that is linked to the sliders to allow the users a visual sense of adjusting their wellbeing as they adjust each metric for their wellbeing. 

This chart in effect functions in the exact opposite fashion of the spiked chart first presented to us wherby the users should here want to maxmise each piece of the chart and try to have their wellbeing represented as a full and even a circle as much as possible. Personally I think this maximisation of wellbeing conveys the goals of the platform succinctly well.

Screenshot 2021-03-09 at 16 05 10

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ gem 'sendgrid-ruby'
 
 gem 'bootstrap', '~> 5.0.0.beta2'
 gem 'font-awesome-sass', '~> 5.15.1'
+gem 'chartkick'
 
 # Use Active Storage variant
 # gem 'image_processing', '~> 1.2'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -82,6 +82,7 @@ GEM
       rack-test (>= 0.6.3)
       regexp_parser (>= 1.5, < 3.0)
       xpath (~> 3.2)
+    chartkick (3.4.2)
     childprocess (3.0.0)
     concurrent-ruby (1.1.8)
     crass (1.0.6)
@@ -265,6 +266,7 @@ DEPENDENCIES
   bootstrap (~> 5.0.0.beta2)
   byebug
   capybara (>= 3.26)
+  chartkick
   database_cleaner-active_record
   devise
   factory_bot_rails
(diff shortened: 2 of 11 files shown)


phil-6 (approved):
Please see below.

phil-6 (commented):
Looks good to me, I can't do any UI testing as yarnkpg appears to be down for me.

I'm happy for this to be merged as a first draft.

Some small improvement ideas. Could the chord of a segment correspond to priority. Could you add border color in to match the rating. On hover could you shade the segment (background color) which is being hovered with its representative colour and display the name of the segment and its word rating somewhere?

My only concern here is that the worse the rating someone gives, the smaller and less visible a section becomes. I think this isn't too big a concern though because the gap it leaves will be obvious.

phil-6:
npm/yarn decided to wake up. 

I definately think coloring / shading would make it clearer. Colours and words are on IJ-34.
I also think it might be worth filling the entire background with the colour, so the negative space when a segment is really small is still indicating the colour. Maybe just with a high transparency value to improve the aesthetic. 
With this it might make sense to remove the numbers on the grid lines as it becomes less of a numeric thing on the front end.

It would also be good to change the number displayed on hover to the word too. 

Hovering over a really small segment is also very hard, is it possible to change the hover area to the area that it could be filling rather than the current slice.
image

IeuanSkinner:
> npm/yarn decided to wake up.
> 
> I definately think coloring / shading would make it clearer. Colours and words are on IJ-34.
> I also think it might be worth filling the entire background with the colour, so the negative space when a segment is really small is still indicating the colour. Maybe just with a high transparency value to improve the aesthetic.
> With this it might make sense to remove the numbers on the grid lines as it becomes less of a numeric thing on the front end.
> 
> It would also be good to change the number displayed on hover to the word too.
> 
> Hovering over a really small segment is also very hard, is it possible to change the hover area to the area that it could be filling rather than the current slice.
> image

The lowest value a user can now give is 1 so there will always be at least a part of the segment on screen. I've removed showing numbers in favour of text labels and added in the colours.

Screenshot 2021-03-10 at 10 28 31

PR #16, opened 2021-03-10
Feature/ij 47 sos button

IeuanSkinner:
I've added an SOS button to the header for users that instantiates a modal window wherein the user can enter the type of situation they're dealing with and any additional info they might have. They can then request help given that information from Include UK team members which involves creating a CrisisEvent record along with sending an SMS and email as post actions.

Screenshot 2021-03-10 at 14 12 47

Currently the sms and email after actions will only print information to the console as this functionality is captured in additional tickets.

Screenshot 2021-03-10 at 14 12 58

### DIFF
--- a/app/controllers/users/dashboard_controller.rb
+++ b/app/controllers/users/dashboard_controller.rb
@@ -3,16 +3,8 @@
 module Users
   # app/controllers/users/dashboard_controller.rb
   class DashboardController < UsersApplicationController
-    before_action :crisis_event, only: :main
-
     def main
       render template: 'users/dashboard/main'
-    end
-
-    private
-
-    def crisis_event
-      @crisis_event = CrisisEvent.new
     end
   end
 end
--- a/app/controllers/users/dashboard_controller.rb
+++ b/app/controllers/users/dashboard_controller.rb
@@ -1,8 +1,18 @@
+# frozen_string_literal: true
+
 module Users
   # app/controllers/users/dashboard_controller.rb
   class DashboardController < UsersApplicationController
+    before_action :crisis_event, only: :main
+
     def main
       render template: 'users/dashboard/main'
+    end
+
+    private
+
+    def crisis_event
+      @crisis_event = CrisisEvent.new
     end
   end
 end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -38,17 +38,51 @@
               <% end %>
             <% end %>
             <% if current_user %>
+              <a class="btn btn-danger m-1" data-bs-toggle="modal" data-bs-target="#sosModal">
+                SOS <i class="fas fa-exclamation-circle"></i>
+              </a>
               <%= link_to edit_user_registration_path, class: "btn btn-success m-1" do %>
                 Edit Profile <i class="fas fa-user-edit"></i>
               <% end %>
               <%= link_to destroy_user_session_path, method: :delete, class: "btn btn-warning m-1" do %>
-                Logout<i class="fas fa-sign-out-alt"></i>
+                Logout <i class="fas fa-sign-out-alt"></i>
               <% end %>
             <% end %>
           </div>
         </div>
       </div>
     </nav>
+
+    <% if current_user %>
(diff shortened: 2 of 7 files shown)


app/controllers/users/dashboard_controller.rb
```diff
@@ -1,8 +1,10 @@
+# frozen_string_literal: true
```
phil-6, line 1: Can tell you have rubocop on now! The frozen string literal is the one rubocop warning I turned off. 

Do you think we need to freeze string literals? If not you can just add
```
Style/FrozenStringLiteralComment:
  Enabled: false
```
to a .rubocop.yml in the root directory

app/views/layouts/application.html.erb
```diff
@@ -38,17 +38,51 @@
...
+                  <%= f.label :additional_info, 'Additional Information'%>
+                </div>
+
+                <%= f.submit 'Request Help', class: 'btn btn-primary mt-2' %>
+              <% end %>
+            </div>
+          </div>
+        </div>
+      </div>
+    <% end %>
```
phil-6, line 85: Do you think its worth extracting modals to separate files and including them with `<%= render '_sos_modal' %>` ?

app/views/layouts/application.html.erb
```diff
@@ -38,17 +38,51 @@
...
+              <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+            </div>
+            <div class="modal-body">
+              <p>If you are in need of help please let us know of the situation you are currently dealing with along
+              with any additional information you can share at this time.</p>
+
+              <%= form_with model: CrisisEvent.new do |f| %>
+                <div class="form-floating mb-2">
+                  <%= f.collection_select :crisis_type_id, CrisisType.all, :id, :category, {}, { class: 'form-select' } %>
+                  <%= f.label :crisis_type_id, 'Situation'%>
```
phil-6, line 71: This is similar to the problems that I was having with @AlexW555 earlier. I think we really want to avoid db / model calls from our views, so ideally these actions should be in a controller.

phil-6: In stead of a drop down select could we do a tick-box style select where each crisis type appears as a select-able button. This should make it easier for people to see the options and make it more touch-friendly.

app/views/layouts/application.html.erb
phil-6: I've also been considering moving the navbar to a partial as well to keep the main application.html.erb file as clean as possible.

db/seeds.rb
```diff
@@ -184,28 +184,28 @@
...
-  team_member_id: 1,
+  category: 'Overdose',
+  team_member_id: 1
 )
 
 CrisisType.create!(
-  category: 'DOMESTIC VIOLENCE',
-  team_member_id: 1,
+  category: 'Domestic Violence',
+  team_member_id: 1
```
phil-6, line 208: If we need to do this elsewhere, ruby does have a `.capitalize` method
https://apidock.com/ruby/String/capitalize

config/routes.rb
```diff
@@ -15,6 +15,8 @@
 
   authenticated :user do
     root 'users/dashboard#main', as: :authenticated_user_root
+
+    post '/crisis_events/new', to: 'users/crisis_events#new', as: :crisis_events
```
phil-6, line 19: if this is a post request should this go to a `new` method or the `create` method?

phil-6 (commented):
Well, you read my mind on a couple of these comments!

app/controllers/users/dashboard_controller.rb
IeuanSkinner: Probably not but they already get added to the generator built controllers though so not including them in controllers we hand craft would be an inconsistency.

config/routes.rb
IeuanSkinner: I've now done this under the feature/IJ-50 branch

app/views/layouts/application.html.erb
IeuanSkinner: I suppose it depends on how many 'Crisis Types' Include UK want, with the 5 we currently have seeded radio buttons will probably be fine but more than that would lead to cluttered controls. Decent follow up question for them I'd say.

phil-6 (approved):

PR #17, opened 2021-03-10
IJ-50

IeuanSkinner:
I've extracted the modals to partials and setup the new crisis event and collection of crisis types as instance variables on the users_application_controller rather than directly accessing the model from the view. I've also introduced an intermediary dialog if a user has created an SOS request in the last hour (and it hasn't been closed).

Screenshot 2021-03-10 at 16 32 47


app/controllers/users/users_application_controller.rb
```diff
@@ -1,6 +1,16 @@
 module Users
   # app/controllers/users/users_application_controller.rb
   class UsersApplicationController < ApplicationController
-    before_action :authenticate_user!
+    before_action :authenticate_user!, :crisis_event, :crisis_types
```
phil-6, line 4: It feels weird to have these running on every action here but as this action is in the navbar is makes sense. Just feels weird.

app/models/user.rb
```diff
@@ -16,6 +16,10 @@ def release
     release_date.present? ? release_date.strftime('%d/%m/%Y') : ''
   end
 
+  def active_crisis_events
+    crisis_events.where('created_at > ? and closed is ?', 1.hours.ago, false).includes(:crisis_type)
```
phil-6, line 20: this is great, exactly what I think they're after, but its definitely worth asking include how they want to handle this behaviour. Please can you add it to the questions section of the requirements doc?

phil-6 (commented):
I feel like the SOS modal could do with a bit more styling. Maybe a red and white border? Or something like that.

the modal names aren't particularly descriptive. I think @bmhtech07 was adding modal as well so maybe update them to `_existing_sos.html.erb` just to avoid any confusion.

app/views/modals/_new.html.erb
```diff
@@ -0,0 +1,28 @@
...
+        <h5 class="modal-title" id="sosModalLabel">SOS Request</h5>
+        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+      </div>
+      <div class="modal-body">
+        <p>If you are in need of help please let us know of the situation you are currently dealing with along
+          with any additional information you can share at this time.</p>
+
+        <%= form_with model: CrisisEvent.new do |f| %>
+          <div class="form-floating mb-2">
+            <%= f.collection_select :crisis_type_id, CrisisType.all, :id, :category, {}, { class: 'form-select' } %>
```
phil-6, line 14: should this be using @crisis_event and @crisis_types now ?

AlexW555 (approved):

IeuanSkinner:
> I feel like the SOS modal could do with a bit more styling. Maybe a red and white border? Or something like that.
> 
> the modal names aren't particularly descriptive. I think @bmhtech07 was adding modal as well so maybe update them to `_existing_sos.html.erb` just to avoid any confusion.

Agreed, I've moved these into /modals/sos so that should provide clarity 👍

app/views/modals/_existing.html.erb
```diff
@@ -0,0 +1,33 @@
...
+              <% else %>
+                No additional information was provided.
+              <% end %>
+              </p>
+            </div>
+          </div>
+        <% end %>
+        <button class="btn btn-primary" data-bs-dismiss="modal" data-bs-toggle="modal" data-bs-target="#sosModal">
+          Create New Request
+        </button>
```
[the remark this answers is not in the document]
IeuanSkinner: Added to doc

app/models/user.rb
IeuanSkinner: Done

app/views/modals/_existing.html.erb
```diff
@@ -0,0 +1,33 @@
...
+              <span><%= crisis_event.crisis_type.category %></span>
+              <span style="float: right"><%= crisis_event.created %></span>
+            </div>
+            <div class="card-body">
+              <p>
+              <% if crisis_event.additional_info.present? %>
+                <%= crisis_event.additional_info %>
+              <% else %>
+                No additional information was provided.
+              <% end %>
```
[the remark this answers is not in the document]
IeuanSkinner: Done

app/views/modals/sos/_new.html.erb
```diff
@@ -15,18 +15,17 @@
...
+              <%= f.collection_radio_buttons :crisis_type_id, @crisis_types_first_half, :id, :category, {} do |b| %>
                 <div class="form-check">
                   <%= b.radio_button(class: 'form-check-input', required: true) + b.label(class: 'form-check-label') %>
                 </div>
               <% end %>
             </div>
 
             <div class="col-md">
-              <%= f.collection_radio_buttons :crisis_type_id, @crisis_types.offset((CrisisType.all.count/2.0).ceil), :id, :category, {} do |b| %>
+              <%= f.collection_radio_buttons :crisis_type_id, @crisis_types_last_half, :id, :category, {} do |b| %>
```
phil-6: This seems like an odd why to do this, but if you're happy with it then :+1:

app/controllers/users/users_application_controller.rb
```diff
@@ -15,10 +15,7 @@ def crisis_event
     end
 
     def crisis_types
-      # @crisis_types = CrisisType.all
-      half_count = (CrisisType.all.count / 2.0).ceil
-      @crisis_types_first_half = CrisisType.all.limit(half_count)
-      @crisis_types_last_half = CrisisType.all.offset(half_count)
+      @crisis_types = CrisisType.all
```
phil-6, line 18: This approach makes a lot more sense to me!

phil-6 (commented):
Functionality looks great. My only thoughts are minor styling considerations.

image

Could we center the "request help" button
Could we change the radio buttons, so each event type is a button. Maybe an outline button when unselected? I feel as though this will make it a bit clearer. 

Overall PR looks great!

app/controllers/users/users_application_controller.rb
IeuanSkinner: Yeah I wrote the limit/offset code at a point where by brain was fried and it didn't even work 😆 that was the bug I was saw in stand-up.

IeuanSkinner:
> Could we center the "request help" button

@phil-6 do the buttons on the existing SOS modal also need to be centered?

phil-6:
Yes, I think for the buttons at the bottom of the modal it makes sense to have them centered.

phil-6 (approved):

PR #20, opened 2021-03-15
Feature/ij 55 journal entries

IeuanSkinner:
This PR includes the addition of 'Journal Entries' which has the same permissions controls that 'WBA Selves' does, the form for creating a new journal entry and the view for displaying journal entries are very bare bones at present as we don't have any sort of wireframe and my focus with this work is to get the functionality in place first.

There's also a sizable refactor here away from the SPA/modal setup as the more I've read in the past week the more I've found that this is not good practice in rails and the dashboard controller in particular ended up being much too large as a result.


config/routes.rb
```diff
@@ -14,24 +14,38 @@
...
-    get '/team_members/:team_member_id/wba/:wba_team_member_id', to: 'team_members/wba_team_members#show', as: :wba_team_member
+      resources :team_members, only: :show do
+        put 'approve', to: 'team_members#approve_team_member', on: :member, as: :approve_team_member
+        put 'admin', to: 'team_members#toggle_admin', on: :member, as: :toggle_admin
+        resources :wba_team_members, path: 'wba', only: :show, as: :wba_team_member
+      end
 
-    get '/users/:user_id', to: 'team_members/users#show', as: :user
+      resources :users, only: :show, as: :user
+    end
```
phil-6, line 48: This is a big restructure. I like the reasoning why but I think we need to work on some aspects of it.
I'm not really happy with the duplicated nested names.

phil-6 (commented):
This is a BIG PR! I've gone though most of the code but haven't actually explored it locally I'll try and do that this afternoon. 
Mostly all good, but a few things which might be worth discussing. I'll let you know when ive had a local play.

app/controllers/team_members/dashboard_controller.rb
```diff
@@ -2,18 +2,18 @@ module TeamMembers
...
-      render template: 'team_members/dashboard/main'
+    def show
+      render 'show'
     end
 
-    private
-
     def unapproved_team_members
       @unapproved_team_members = TeamMember.all.where(approved: false).order(:id)
     end
```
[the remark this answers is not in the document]
IeuanSkinner: It's a weird one because making it private doesn't seem to break anything but the IDE was complaining about not being able to find the action.

config/routes.rb
IeuanSkinner: Which duplicates?

app/views/users/journal_entries/new.html.erb
```diff
@@ -0,0 +1,38 @@
...
+            </div>
+            <div class="col-12 text-center">
+              <%= f.submit 'Submit', class: 'btn btn-primary mt-2' %>
+            </div>
+          <% end %>
+        </div>
+      </div>
+    </div>
+  </section>
+</main>
```
phil-6, line 38: This needs much more consideration for mobile. Smaller margins and padding. Label/Placeholder could do with be more descriptive as well. 

image

app/views/users/journal_entry_permissions/new.html.erb
```diff
@@ -0,0 +1,40 @@
...
+                  </div>
+                </div>
+              </div>
+            <% end %>
+          </div>
+        </div>
+      </div>
+    </div>
+  </section>
+</main>
```
phil-6, line 40: I feel as though this doesn't make it obvious which options have been selected. Maybe add a Font-Awesome tick and cross which swap when selected or not. 
image

phil-6 (commented):
There is an amazing amount of progress in this PR. Great work. 
While I think there are important changes which need to be made here I'm in two minds about whether they should be made to this PR or if this should be merged and the improvements spun out into their own tickets. 

I feel as though we are losing a lot from the dashboard by splitting everything out into its own view. I think partials might be a better approach for this so we can still have the "dashboard" .
 
I'd also be interested in exploring some form of transition for the turbolinks - which hopefully would make everything feel a bit more connected. 

We've also lost the button to add a new wba.

IeuanSkinner:
> We've also lost the button to add a new wba.

I can't see this being lost 🤔  you can only fill out 1 WBA per day now is that what you're seeing?

phil-6:
> > We've also lost the button to add a new wba.
> 
> I can't see this being lost thinking you can only fill out 1 WBA per day now is that what you're seeing?

Yep, that'll be it. Probably need to make this clearer! :laughing:

app/views/users/dashboard/show.html.erb
```diff
@@ -120,87 +120,4 @@
       </div>
     </div>
   </section>
```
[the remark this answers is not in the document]
IeuanSkinner: We'll need to discuss this then as there's no straightforward way of taking that approach and if we're going to defy conventions and take that approach there's lots of considerations like will others be able to contribute effectively and/or could we hand this off effectively.

app/javascript/packs/permissions.js
```diff
@@ -0,0 +1,23 @@
...
+          icon = checkbox.closest('div').querySelector('i')
+
+    if (checkbox.checked) {
+        label.style['background-color'] = '#70A739'
+        label.style['border-color'] = '#70A739'
+        icon.classList.add('fa-check-circle')
+        icon.classList.remove('fa-times-circle')
+    } else {
+        label.style['background-color'] = '#F32516'
+        label.style['border-color'] = '#F32516'
```
phil-6, line 14: Could these be imported from the CSS? Or reference they style sheet?

phil-6 (approved):
I'm happy for this to be merged. I think there might be a few bits which we revisit later but it looks good to me for now. Great work!

app/javascript/packs/permissions.js
IeuanSkinner: No mechanism I know for doing this, does rails have something?

IeuanSkinner: Oh I suppose I could use a class name and toggle that rather than direct style rules.

PR #21, opened 2021-03-16
Feature/ij 57 admin manage crisis

IeuanSkinner:
Team members can now see all of the Crisis Event records that have been created from their scope via the 'Crisis Events' button on the homepage. This will take them to the index view which lists only the active crisis events with the option of viewing the closed crisis events via a button which will take the user to a paginated view for all of the closed crisis events sorted by creation date newest to oldest.

A team member can click on a crisis event in either the active crisis events index or the paginated closed crisis events index and be taken to the crisis event member record where they can see the full details of the crisis events along with the option to close the crisis event and add notes to the crisis event.

Crisis events button on homepage
Screenshot 2021-03-16 at 14 30 35

Active crisis events index
Screenshot 2021-03-16 at 14 30 48

Paginated closed crisis events index
Screenshot 2021-03-16 at 14 30 54

Crisis event member view
Screenshot 2021-03-16 at 14 31 11

### DIFF
--- a/app/assets/stylesheets/crisis_events.scss
+++ b/app/assets/stylesheets/crisis_events.scss
@@ -15,6 +15,18 @@
           float: right;
         }
       }
+
+      .card-footer {
+        .col-6 {
+          &:first-child {
+            text-align: left;
+          }
+
+          &:last-child {
+            text-align: right;
+          }
+        }
+      }
     }
   }
 
--- a/app/assets/stylesheets/crisis_events.scss
+++ b/app/assets/stylesheets/crisis_events.scss
@@ -1,3 +1,20 @@
 // Place all the styles related to the CrisisEvents controller here.
 // They will automatically be included in application.css.
 // You can use Sass (SCSS) here: https://sass-lang.com/
+
+@import "variables";
+
+.team_members {
+  &.crisis_event {
+    .card {
+      text-align: left;
+      color: $black;
+
+      .card-header {
+        span:last-child {
+          float: right;
+        }
+      }
+    }
+  }
+}
--- a/app/assets/stylesheets/crisis_events.scss
+++ b/app/assets/stylesheets/crisis_events.scss
@@ -17,4 +17,20 @@
       }
     }
   }
+
+  &.crisis_events {
+    .controls {
+      padding: 0;
+
+      .col {
+        &:first-child {
+          text-align: left;
+        }
+
+        &:last-child {
+          text-align: right;
+        }
+      }
+    }
+  }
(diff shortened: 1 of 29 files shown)


phil-6 (changes requested):
Code looks good however some things don't appear to be working fully on the front end.

Closed Crisis Events page
image
Could also do with a link back to active events.
Replace last update column with closed_at

Show Crisis Event
image
Could do with a "back to list" button.
Change "close" to "close event"
If "Additional Info" is empty then display "User has not provided any additional info"
update time to include time since event (eg 35 minutes ago, or a line which says something like "Nadia Will hit the SOS button 35 minutes ago")


It would be nice to have an "edit note" button, (maybe time limited for 5 minutes) to allow of correcting of spelling mistakes / typos etc.

It would be good to have the "closed" at date displayed on closed crisis events.

Active crisis events page is now being weird:
image
image

It would be good to update tables for mobile so their overflow is scrollable instead of expanding the entire viewport. I've done this on SP like this:
```scss  
table {
    overflow-x: auto;
    white-space: nowrap;
    display: block;
    @media (min-width: 576px) {
      display: table;
    }
    ...
}
```

On mobile remove x padding from crisis_event container, and x margin from row to make better use of limited horizontal space.

Notes don't work on mobile.
image

We also want to display the users phone number on the event info.

IeuanSkinner:
@phil-6 I believe I've addressed everything you raised, I haven't been able to recreate the issues you were seeing on the tables with that weird text overlay but I have fixed it so table cells don't wrap and use the ellipsis css rule so I think it should still be addressed. If not let me know and we can take a closer look.

For editing the crisis event notes I've left that for now as it's something I'd like to run it past Include UK.

phil-6 (changes requested):
Looking a lot better but still a few bugs when moving around the UI.  I really like the mobile design changes and the back button is a great addition. 
- After adding a note to a closed crisis event the back button appears to stop working. This actually appears to occur after adding any note.
- Back button also doesn't work after closing a crisis event.
- Simple formatting is not preserved in notes. (wrap outputs in simple_format() - https://apidock.com/rails/ActionView/Helpers/TextHelper/simple_format

It might be good hard code the back button paths. 

I can also no longer recreate the weird text overlay bug from before either. I suspect TurboLinks had a moment and hadn't loaded in correctly.

IeuanSkinner:
@phil-6 can you take another look at this? If happy I can merge this and then IJ-73

phil-6 (approved):
Looks good to me now.

PR #22, opened 2021-03-16
IJ-73 WBA Self View Logs are now in a paginated view.

IeuanSkinner:
The WBA Self View Logs records are no longer rendered in a single large table on the team member page but are now in their own seperate paginated view.

Screenshot 2021-03-16 at 16 25 21

Screenshot 2021-03-16 at 16 25 25

### DIFF
--- a/app/controllers/team_members/wba_self_view_logs_controller.rb
+++ b/app/controllers/team_members/wba_self_view_logs_controller.rb
@@ -1,0 +1,55 @@
+module TeamMembers
+  # app/controllers/team_members/wba_self_view_logs_controller.rb
+  class WbaSelfViewLogsController < AdminApplicationController
+    before_action :team_member, only: %i[index page_index]
+    before_action :page, :offset, :wba_self_view_logs, only: :page_index
+    before_action :count, :last_page, only: :page_index, if: -> { @wba_self_view_logs.present? }
+    before_action :redirect, only: :page_index, unless: -> { @wba_self_view_logs.present? }
+
+    LIMIT = 5
+
+    # GET /team_members/:team_member_id/wba_self_view_logs
+    def index
+      redirect_to pages_team_member_wba_self_view_logs_path(@team_member, 1)
+    end
+
+    # GET /team_members/:team_member_id/wba_self_view_logs/page/:page_number
+    def page_index
+      render 'index'
+    end
+
+    private
+
+    def count
+      @count = @team_member.wba_self_view_logs.count
+    end
+
+    def last_page
+      @last_page = @offset + LIMIT >= @count
+    end
+
+    def offset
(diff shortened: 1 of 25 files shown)


app/controllers/team_members/wba_self_view_logs_controller.rb
```diff
@@ -0,0 +1,55 @@
+module TeamMembers
+  # app/controllers/team_members/wba_self_view_logs_controller.rb
+  class WbaSelfViewLogsController < AdminApplicationController
+    before_action :team_member, only: %i[index page_index]
+    before_action :page, :offset, :wba_self_view_logs, only: :page_index
+    before_action :count, :last_page, only: :page_index, if: -> { @wba_self_view_logs.present? }
+    before_action :redirect, only: :page_index, unless: -> { @wba_self_view_logs.present? }
+
+    LIMIT = 5
```
phil-6, line 9: I still think "LIMIT" could be more descriptively named.

phil-6 (commented):
Looks all good. I've mentioned a few bits on the Journal Entries PR which also apply here. 

I think it would be useful to add some extra bits to the paginated views:
- page indicator at the bottom, which shows something along the lines of 1...3 *4* 5 ... 20, or 'first' 3 *4* 5 'last'. 
- total number of items in the collection
- change 'created at' to 'viewed at'
- add 'created at' which relates to the item which was viewed and provides a link to the item 
- it would be cool to allow filtering by user - but that might be worth creating a new ticket for.

IeuanSkinner:
>     * total number of items in the collection
> 
>     * change 'created at' to 'viewed at'
> 
>     * add 'created at' which relates to the item

@phil-6 I've added the above as they're simple additions. The remaining asks are new criteria so should be added to a new ticket as future work.

phil-6:
Ace, thanks Ieuan. Please can you create the new tickets?

app/controllers/team_members/wba_self_view_logs_controller.rb
```diff
@@ -0,0 +1,55 @@
+module TeamMembers
+  # app/controllers/team_members/wba_self_view_logs_controller.rb
+  class WbaSelfViewLogsController < AdminApplicationController
+    before_action :team_member, only: %i[index page_index]
+    before_action :page, :offset, :wba_self_view_logs, only: :page_index
+    before_action :count, :last_page, only: :page_index, if: -> { @wba_self_view_logs.present? }
+    before_action :redirect, only: :page_index, unless: -> { @wba_self_view_logs.present? }
+
+    LOGS_PER_PAGE = 5
```
phil-6, line 9: I'm much happier with this, thanks.

phil-6 (approved):

PR #24, opened 2021-03-18
Feature/ij 75 journal entries view

IeuanSkinner:
Team members can now view a paginated index of all the journal entries that have been created **that they have an associated permissions record for.** Once a team member then views the journal entry a view log is created and this is observable for them so they can keep track of what they have/haven't read.

The full index of these view logs can be read by admins by looking at the team members member view as is the case for the WBA Self View Logs records.

Team member dashboard includes 'Journal Entries' button that links to paginated index.
Screenshot 2021-03-17 at 17 32 08

Index shows a team member journal entries they are permitted to see through the journal entry permissions record. Viewed status is also shown.
Screenshot 2021-03-17 at 17 32 14

Journal entry member record
Screenshot 2021-03-17 at 17 32 20

This PR also includes the abstractions for the controllers I've documented on LucidChart.


phil-6 (approved):
LGTM :+1:  :smile:

PR #28, opened 2021-03-22
Feature/ij 80 team member pinned users

IeuanSkinner:
I've added in the ability for team members to pin and unpin users on the user index view, they can also rearrange these pinned users via the chevron arrows next to the pin icon.

Currently I haven't put a limit on the number of users a team member can pin as any amount would be arbitrary and there's no reason why they shouldn't be able to pin as many as they'd like.

Screenshot 2021-03-22 at 13 19 32


phil-6 (approved):
LGTM, some good other little improvements here too.

AlexW555:
Functions perfectly, not too sure if this is possible, but when users are pinned could the pin icon be highlighted or changed colour?  

https://user-images.githubusercontent.com/17786854/112153937-b7a3dc00-8bdb-11eb-8924-b98b1c24fd83.mov

PR #29, opened 2021-03-22
Ij 72 signup terms

AlexW555:
Added a small terms page. The information on it for now is pretty generic. It's main purpose is to fix a broken link and add a basic page which covers key areas which users would want to see. Including: general use terms, cancellation and links to external sites.


app/views/pages/terms.html.erb
```diff
@@ -0,0 +1,139 @@
...
+      <h1>Your Include Journey</h1>
+      <p>track and progress on your journey to make the changes you want and need.</p>
+      <p>Built for Include UK by the Legal Innovation lab Wales</p>
+      <div class="text-center">
+        <%= link_to new_team_member_session_path, class: 'btn btn-secondary' do %>
+          team member sign in <i class="fas fa-users"></i><i class="fas fa-sign-in-alt"></i>
+        <% end %>
+        <%= link_to new_user_session_path, class: 'btn btn-success' do %>
+          user sign in <i class="fas fa-user"></i><i class="fas fa-sign-in-alt"></i>
+        <% end %>
```
IeuanSkinner, line 13: Terms isn't in the unauthenticated block so it's available in platform at /terms for all users (signed in and not signed in). 

If this is intentional these links need to come out when a user is authenticated.

app/views/pages/terms.html.erb
```diff
@@ -0,0 +1,139 @@
...
+      <div class="col-xs-12">
+        <h2>Terms and conditions</h2>
+        <p>
+          Include UK operates Include Journey (as defined below), your use of Include Journey is
+          governed by these Terms of Service (“Terms”), and your use of Include Journey constitutes your
+          acceptance
+          of the Terms. The application was designed and developed by the legal Innovation Lab Wales.
+        </p>
+        <p>
+          "Include Journey" (the "Service") means Include Journey project solution.
```
IeuanSkinner, line 26: I don't understand this line. Are we going to get the terms written by Include or some other legal person? It should never be on programmers to do this sort of thing.

AlexW555: Sorry, This was just taken from a generator site that advertised itself for projects just needing a terms page. I know it isn't meant to be fully written up. As this isn't live, I just copied what it generated to build up a page to get an idea of what it would look like. With some relevant information that wasn't full of Lorem. I'll just replace the sections now

app/views/pages/terms.html.erb
AlexW555: As this page is defined as "Sign up terms",  moved the route to unauthenticated.

IeuanSkinner:
So adding the terms route to the unauthenticated block isn't enough as there's no equivalent route overwriting that in the authenticated blocks. I've added that in as a  redirect to route.

I've also updated the text under use as currently we don't do age verification nor do we require approval for sign up for users, that's only a constraint on team members.

IeuanSkinner (approved):
I'm happy with this as-is for now @AlexW555 can you create a ticket on Jira to get the terms of use worked on by someone with expertise in this area?

phil-6:
Shouldn't authenticated users still be able to view the terms page?

IeuanSkinner:
> Shouldn't authenticated users still be able to view the terms page?

@phil-6 https://github.com/Legal-Innovation-Lab-Wales/include-journey/pull/29#discussion_r598882043

phil-6:
Can we just move it outside of the authorisation blocks so it is globally accessible?

AlexW555:
> Can we just move it outside of the authorisation blocks so it is globally accessible?

I can do, to make this page globally accessible it would mean these sign-in buttons would have to be removed when an authrenticated user visits and replaced by a "go back" button. 

As these are terms and conditions of signing up I agree it should be accessible to users who have already signed up. I just didn't know if adding a "go back" button taking the user out of the "authenticated" route could cause issues
Screenshot 2021-03-23 at 1 45 59 pm

phil-6:
@AlexW555  I don't think we need sign up buttons on this page. 
If a person is signed in they should have access to the nav bar.

The most likely time someone will see this is during the sign up process, where it should be a simple page to give them the information they're looking for as easily as possible.

AlexW555:
View for current users signed in
Screenshot 2021-03-24 at 3 11 48 pm
View for unauthenticated users
Screenshot 2021-03-24 at 3 12 00 pm

IeuanSkinner (approved):

phil-6 (approved):

PR #30, opened 2021-03-22
IJ-81 Team members can now search for a given user (first name/last n…

IeuanSkinner:
Team members can now search for a given user (first name/last name) and the index view will update to display just this user.

Screenshot 2021-03-22 at 15 55 13


phil-6 (approved):
LGTM

PR #32, opened 2021-03-23
IJ-79 Navbar button animations

IeuanSkinner:
Updated the buttons to use darker colour as they're sitting on top of a light background, also added in transitions for smoother feel when revealing text.

https://user-images.githubusercontent.com/6815945/112133260-6daefc00-8bc3-11eb-8099-6840948a0803.mp4


phil-6 (approved):
Nice, I like this method. Works well. Good work :grinning:

AlexW555 (approved):
👍🏻

PR #33, opened 2021-03-23
Alerts to notice update

IeuanSkinner:
Where a message is not a warning but rather just information to be shown to a user I've changed the type from alert to notice.

### DIFF
--- a/app/controllers/users/journal_entries_controller.rb
+++ b/app/controllers/users/journal_entries_controller.rb
@@ -17,7 +17,7 @@ module Users
     # POST /journal_entries
     def create
       if (@journal_entry = current_user.journal_entries.create!(journal_entry_params))
-        redirect_to new_journal_entry_permission_path(@journal_entry), notice: 'New journal entry added'
+        redirect_to new_journal_entry_permission_path(@journal_entry), success: 'New journal entry added'
       else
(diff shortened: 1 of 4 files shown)


app/controllers/users/journal_entries_controller.rb
```diff
@@ -17,7 +17,7 @@ def dashboard
     # POST /journal_entries
     def create
       if (@journal_entry = current_user.journal_entries.create!(journal_entry_params))
-        redirect_to new_journal_entry_permission_path(@journal_entry), alert: 'New journal entry added'
+        redirect_to new_journal_entry_permission_path(@journal_entry), notice: 'New journal entry added'
```
phil-6, line 20: this could be 'success:'

app/controllers/users/users_application_controller.rb
```diff
@@ -22,7 +22,7 @@ def permissions_required
       return unless current_user.permissions_required?
 
       redirect_to new_journal_entry_permission_path(current_user.journal_entries.last),
-                  alert: 'Please specify which team members can view this journal entry'
+                  notice: 'Please specify which team members can view this journal entry'
```
phil-6, line 25: I think this one should still be `alert:` as its a warning message.

phil-6 (approved):
Thanks, LGTM

PR #37, opened 2021-03-23
IJ-83 Added home page routes for authenticated users.

IeuanSkinner:
Screenshot 2021-03-23 at 16 36 12

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,9 +1,11 @@
+# app/controllers/pages_controller.rb
 class PagesController < ApplicationController
   def main
     if session[:awaiting_approval_notice].present?
       flash.now[:alert] = session[:awaiting_approval_notice]
       session.delete(:awaiting_approval_notice)
     end
+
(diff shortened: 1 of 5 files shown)


phil-6 (commented):
Could we not have a single route outside of the authentication scopes which just does `get 'home', to: 'pages#home'`? Or did that not work?

IeuanSkinner:
> Could we not have a single route outside of the authentication scopes which just does `get 'home', to: 'pages#home'`? Or did that not work?

It won't work because "authenticated" team members still require approval before we let them do anything in system (rightly) so we need to make sure that check is still happening as a super-action to everything else. For users we need to setup the instance variables used by the crisis events modal window which we want to still be accessible from this page.

Both of these are UserApplicationController actions so having a home function there and routing each respectively to it actually makes the most sense.

phil-6:
Ahhh, of course. 

@AlexW555 this pattern is what you'll have to follow for the terms page as well.

phil-6 (approved):

PR #38, opened 2021-03-24
Feature/ij 85 pagination updates

IeuanSkinner:
The pagination controller has been converted to a partial as I was reusing that HTML segment in multiple places, I've also now added a number line to the pagination controller and I've also turned the user search field into a partial so I can reuse it between the team member user index and team member journal entries view logs index as seen below.

Screenshot 2021-03-24 at 13 05 39

### DIFF
--- a/app/controllers/team_members/users_controller.rb
+++ b/app/controllers/team_members/users_controller.rb
@@ -88,7 +88,7 @@ module TeamMembers
     end
 
     def user_count
-      @user_count = User.all.count
+      @user_count = User.count
     end
 
     def user_pin
--- a/app/controllers/team_members/users_controller.rb
+++ b/app/controllers/team_members/users_controller.rb
@@ -6,8 +6,7 @@ module TeamMembers
     before_action :verify_pin, only: :pin
     before_action :verify_unpin, only: :unpin
     before_action :query, :pinned_users, :active_users, :user_count, only: :index
-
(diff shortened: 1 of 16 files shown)


phil-6:
I've just added [IJ-113](https://lilw.atlassian.net/browse/IJ-113) which might be addressed as part of this.  (Added ticket before I remember this was open.

phil-6:
[IJ-115](https://lilw.atlassian.net/browse/IJ-115) Could also be addressed in this ticket, or if its too complex could be addressed on its own later.

IeuanSkinner:
@phil-6 we're better off addressing it seperately as it wasn't part of the criteria for this work.

app/controllers/team_members/users_controller.rb
```diff
@@ -86,6 +87,10 @@ def user
       @user = User.find(params[:id])
     end
 
+    def user_count
+      @user_count = User.all.count
```
phil-6, line 91: I think this can be simplified to `User.count`

phil-6 (commented):
Please can we update the styling to remove the underline on the page controls:
image

other than that, it all looks good to me.

IeuanSkinner:
@phil-6 I've added the changes you've asked for, anything more?

phil-6 (approved):
Great!

PR #39, opened 2021-03-24
Journal entry permissions change

IeuanSkinner:
Removed the requirement that at least one permission needs to be set on a journal entry so users can now submit journal entries only viewable by themselves. Also fixed that journal entries in the team member index were not latest first.


phil-6 (approved):

AlexW555 (approved):
👍🏻

PR #41, opened 2021-03-24
IJ-17 - Includes: IJ-39 IJ-38 IJ-46 IJ-131 IJ-132 IJ-133

bmhtech07:
Formatted view for for 'dashboard' style view and copied sos format from TeamMembers Users Index.

Still to test using oneliner for IP location, to display next to user 'last signed in at' (added to Jira backlog).

Left placeholder for Wellbeing History chart / display.

Intuitions says we should break out of the global lowercase for the user's name in the H2 at top.

image


app/controllers/team_members/notes_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    # GET /notes or /notes.json
+    def index
+      @notes = Note.user
+    end
+
+    # GET /notes/1 or /notes/1.json
+    def show; end
+
+    # GET /notes/1/edit
+    def edit; end
```
IeuanSkinner, line 17: These can be removed

app/controllers/team_members/notes_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    # POST /notes or /notes.json
+    def create
+      @note = Note.create!({ content: note_params[:content],
+                             visible_to_user: note_params[:visible_to_user],
+                             user: @user,
+                             team_member: current_team_member })
```
IeuanSkinner, line 24: As we have the user setup via `@user` we can create this note against that users notes relationship and so we can save a parameter i.e.

```
@note = @user.notes.create!(content: note_params[:content], 
                                                 visible_to_user: note_params[:visible_to_user],
                                                 team_member: current_team_member)
```

or we could set it up against the team member and save a parameter i.e.

```
@note = current_team_member.notes.create!(content: note_params[:content], 
                                                                            visible_to_user: note_params[:visible_to_user],
                                                                            user: @user)
```

app/controllers/team_members/notes_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      redirect_to user_path(@user), notice: @note ? 'Note added!' : "We couldn't create the note. Please try again."
```
IeuanSkinner, line 26: Rather than personifying the warning keep it robotic i.e. `Note could not be created. Please try again.`

app/controllers/team_members/notes_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    # PATCH/PUT /notes/1 or /notes/1.json
+    def update
+      redirect_to user_path(@user), notice: @note.update!(note_params) ? 'Note updated!' : "We couldn't create the note. Please try again."
+    end
+
+    # DELETE /notes/1 or /notes/1.json
```
IeuanSkinner, line 34: These references to `.json` variants of the URL can be removed.

app/controllers/team_members/notes_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    def destroy
+      @note.destroy
+      redirect_to notes_url, notice: 'Note was successfully destroyed.'
+    end
+
+    private
+
+    # Use callbacks to share common setup or constraints between actions.
```
IeuanSkinner, line 42: Strip this out.

app/controllers/team_members/notes_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    def set_note
+      @note = Note.find(params[:id])
+    end
+
+    def user
+      @user = User.find(params[:user_id])
+    end
+
+    # Only allow a list of trusted parameters through.
```
IeuanSkinner, line 51: Also strip this out.

app/controllers/team_members/notes_controller.rb
```diff
@@ -0,0 +1,57 @@
+module TeamMembers
+  # app/controllers/team_members/notes_controller.rb
+  class NotesController < TeamMembersApplicationController
+    before_action :set_note, only: %i[show edit update destroy]
+    before_action :note_params, only: :create
+    before_action :user
+
+    # GET /notes or /notes.json
+    def index
+      @notes = Note.user
```
IeuanSkinner, line 10: This is quite odd as it's creating a `@notes` instance variable but it's being bound to a `User` model? I'm not even sure what user model it would be given the note doesn't appear to be a specific instance (no `.find(some_id)`). Is this actually being called/used or is it some leftover code? My hunch is the latter.

app/javascript/packs/application.js
```diff
@@ -6,8 +6,8 @@
 import Rails from "@rails/ujs"
 import Turbolinks from "turbolinks"
 import * as ActiveStorage from "@rails/activestorage"
-import "bootstrap"
 import "channels"
+import "bootstrap"
```
IeuanSkinner, line 10: Was there a reason behind swapping these import orders?

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+    <div class="col-12 col-md-6 mb-4">
+      <div class="card h-100" id="users">
+        <div class="card-header">
+          <h3>User Notes</h3>
+        </div>
+        <div class="card-body bg-transparent notes-card" style="max-height: 40vh; overflow: scroll;">
+          <% if @user_notes.empty? %>
+            <span>There are no notes on this user yet.</span>
+          <% end %>
+          <% unless @user_notes.empty? %>
```
IeuanSkinner, line 206: If followed by unless can be replaced with an if-else, that way the `@user_notes.empty?` evaluation only needs to run once i.e.

```
<% if @user_notes.empty? %>
...
<% else %>
...
<% end %>

File: app/views/team_members/users/show.html.erb [hunk 10]
```diff
@@ -0,0 +1,236 @@
...
+                <div class="card-header d-flex justify-content-between">
+                  <span class="me-4">Added by <strong><%= note.team_member.first_name %></strong> on <strong><%= note.created %></strong></span>
+                  <span>Visible to user?
+                    <% if note.visible_to_user? %><i class="fa fa-check-circle text-success ms-2"></i>
+                    <% else %> <i class="fa fa-times-circle ms-2 text-danger"></i>
+                    <% end %>
+                  </span>
+                </div>
+                <div class="card-body text-start">
+                  <p><%= note.content %></p>
```
IeuanSkinner, line 218: Wrap in `simple_format` to preserve line breaks.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
+<main class="container team-members users">
+  <section class="row">
+    <div class="col">
+      <h2 class="my-3">User: <%= @user.full_name %></h2>
+    </div>
+  </section>
+  <!-- User bio / details section -->
```
IeuanSkinner, line 7: These markup comments can be removed.

app/controllers/team_members/notes_controller.rb
phil-6: What are your thoughts behind this Ieuan? Why robotic over humanised?

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+                    </div>
+                  <% end %>
+                  <% unread_journal_entries = current_team_member.unread_journal_entries(@user) %>
+                  <% unless unread_journal_entries.zero? %>
+                    <div class="row">
+                      <%= link_to '#' do %>
+                        <i class="fas fa-book-reader"></i> <%= unread_journal_entries %> unread journal entries
+                      <% end %>
+                    </div>
+                    <% unless active_crisis_events.empty? %>
```
IeuanSkinner, line 47: This was chopped and swapped from the user index page and as such has the same bug, this block for active crisis is inside the journal entries block so if the user doesn't have any unread journal entries then their active crisis count won't appear. This needs to be moved outside of this block.

app/controllers/team_members/users_controller.rb
```diff
@@ -53,6 +54,40 @@ def resources
...
+    def user_location
+      @user_location = Timeout::timeout(5) { Net::HTTP.get_response(URI.parse('http://api.hostip.info/country.php?ip=' + @user.last_sign_in_ip )).body } rescue "Unknown"
+    end
+
+    def wellbeing_metrics
+      @wellbeing_metrics = WellbeingMetric.all
+    end
+
+    def wba_self
+      @wba_self = WbaSelf.includes(:wba_self_scores).find(params[:id])
```
IeuanSkinner, line 78: This isn't going to fetch the last wellbeing assessment, it's going to fetch the first wellbeing assessment given the user ID. 

We have the user instantiated as `@user` and the `User` model has a function for fetching the last wellbeing assessment a user completed.

app/controllers/team_members/notes_controller.rb
phil-6: this also looks like its only used in the `notes/index.html.erb` page, which also appears to be somewhat borked. I'd strip it out for now. 
I'd agree with Ieuan's hunch

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+            <div class="col-12 col-lg-5 text-start">
+              <h4>Current wellbeing</h4>
+              <span><strong>Last updated: </strong><%= @wba_self.created %></span>
+              <canvas class="mt-4 wellbeing-chart" id="wellbeing-chart" width="400" height="400"></canvas>
+              <div class="col-md-4 sliders d-none">
+                <% @wellbeing_metrics.each do |wellbeing_metric| %>
+                  <label class="hidden" for=<%= "wellbeing_metric_#{wellbeing_metric.id}" %>><%= wellbeing_metric.name %></label>
+                  <input class="hidden" type="range" id="<%= "wellbeing_metric_#{wellbeing_metric.id}" %>"
+                         value="<%= @wba_self.wba_self_scores.find_by(wellbeing_metric_id: wellbeing_metric.id).value %>">
+                <% end %>
```
IeuanSkinner, line 119: This code has a bug as it iterates over the wellbeing metrics, attempts to find the corresponding score for the wellbeing assessment and gets the `.value` from it but if a new wellbeing metric has been added since the assessment was done the `find_by` will evaluate to `nil` as so calling `.value` on it will throw an exception.

To better fetch this information see the latest code for the user index `_card.html.erb`. We can pull through the scores from the wellbeing assessment and pull the wellbeing metrics from those scores rather than the other way around and so if a new wellbeing metric is added then there's no harm as there's just no score record for that wellbeing assessment that maps to that metric as expected.

app/views/team_members/users/show.html.erb
phil-6: I don't fully hate them in a big HTML file - but the section should be clear from the headers and classes

app/controllers/team_members/users_controller.rb
```diff
@@ -53,6 +54,40 @@ def resources
...
+    rescue ActiveRecord::RecordNotFound
+      session notice: 'No wellbeing assessment could be found'
+    end
+
+    def active_crisis
+      @active_crisis = @user.crisis_events.active
+    end
+
+    def journal_entries
+      @journal_entries = @user.journal_entries
```
IeuanSkinner, line 88: You're pulling the journal entries directly from the user here which completely subverts the permissions structure we have in place. You need to get the journal entries from the team member model where the `user_id` on the journal entry matched this `@user`.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+        </div>
+        <div class="card-body">
+          <% if @journal_entries.count > 0 %>
+            <h4>Latest journal entry</h4>
+            <div class="card bg-transparent px-0 my-3">
+              <div class="card-header px-0">
+                <span class="mx-3"><strong>Entered on: </strong></span><%= @journal_entries.last.created %>
+              </div>
+              <div class="card-body text-start">
+                <p><%= @journal_entries.last.entry %></p>
```
IeuanSkinner, line 148: Here you're displaying a journal entry without adding a view log for the team member meaning a team member can read a users journal entries via this page and an admin would never know, I suggest not showing the entry anywhere other than the journal entry show view.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+              </div>
+            </div>
+          <% else %>
+            <span>The User hasn't completed any Journal entries yet</span>
```
IeuanSkinner, line 152: Mixed casing.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+          <% unless @active_crisis.empty? %>
+              <p><span class="h5">When? </span><%= @active_crisis.first.created %></p>
+              <p><span class="h5">Type? </span> <%= @active_crisis.first.crisis_type.category %></p>
+              <p><span class="h5">Info: </span> <%= @active_crisis.first.additional_info %></p>
+          <% end %>
+          <% if @active_crisis.empty? %>
+            <p class="text-start">
+              This user has no active Crisis Events. Any active Crisis Events will be displayed here.
+            </p>
+          <% end %>
```
phil-6, line 92: there are 4 logic evaluations here, I feel as though it could be reduced to one if..else block.
`if @active_crisis.present?` etc

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+          </button>
+        </div>
+      </div>
+    </div>
+  </div>
+  <!-- End Notes Section -->
+</main>
+
+<%= javascript_pack_tag 'polar_area_chart' %>
+<%= javascript_pack_tag 'line_chart' %>
```
phil-6, line 236: This is throwing a JS error, I'd suggest we remove the pack tag until we have it working

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+          <% if @user_notes.empty? %>
+            <span>There are no notes on this user yet.</span>
+          <% end %>
+          <% unless @user_notes.empty? %>
+            <% @user_notes.each do |note| %>
+              <div class="card bg-transparent my-2">
+                <div class="card-header d-flex justify-content-between">
+                  <span class="me-4">Added by <strong><%= note.team_member.first_name %></strong> on <strong><%= note.created %></strong></span>
+                  <span>Visible to user?
+                    <% if note.visible_to_user? %><i class="fa fa-check-circle text-success ms-2"></i>
```
IeuanSkinner, line 212: Have the conditional statement be on one line and the result of that conditional evaluating to true on the next.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+              <p><span class="h5">Info: </span> <%= @active_crisis.first.additional_info %></p>
+          <% end %>
+          <% if @active_crisis.empty? %>
+            <p class="text-start">
+              This user has no active Crisis Events. Any active Crisis Events will be displayed here.
+            </p>
+          <% end %>
+        </div>
+        <div class="card-footer py-3">
+        </div>
```
phil-6, line 95: Remove this as we're not using the footer.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+    <!-- Notes section -->
+    <!-- Add note modal -->
+    <div class="modal fade" id="note-modal" tabindex="-1" role="dialog" aria-labelledby="note-modal" aria-hidden="true">
+      <div class="modal-dialog modal-lg" role="document">
+        <div class="modal-content">
+          <%= form_with model: @note,
+                        url: user_add_note_index_path(@user),
+                        method: :post do |form| %>
+            <div class="modal-header border-0">
+              <h5 class="modal-title text-black-50" id="">Add a note for this user</h5>
```
IeuanSkinner, line 174: Blank ID

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+      </div>
+    </div>
+    <!-- End modal -->
+
+    <div class="col-12 col-md-6 mb-4">
+      <div class="card h-100" id="users">
+        <div class="card-header">
+          <h3>User Notes</h3>
+        </div>
+        <div class="card-body bg-transparent notes-card" style="max-height: 40vh; overflow: scroll;">
```
IeuanSkinner, line 202: Remove inline style, style rule should be in sass. Also having overflow set to scroll will have scrollbars always appearing event without notes.

Screenshot 2021-03-25 at 08 52 55

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
+<main class="container team-members users">
+  <section class="row">
+    <div class="col">
+      <h2 class="my-3">User: <%= @user.full_name %></h2>
```
IeuanSkinner, line 4: I'd remove this `User: ` prefix, this page already says user in a lot of places.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
+<main class="container team-members users">
+  <section class="row">
```
IeuanSkinner, line 2: Needs a 'Back to users' button like we have on many other views that directs a team member to the user index page.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+    <div class="col-12 col-md-6 mb-4">
+      <div class="card h-100 <%= @user.crisis_events.active.empty? ? '' : 'has-crisis' %>">
+        <div class="card-header">
+          <% unless @active_crisis.empty? %>
+            <h3>SOS status:
+              <span class="h2">Crisis </span><i class="fas fa-exclamation-triangle"></i>
+            </h3>
+          <% end %>
+          <% if @active_crisis.empty? %>
+            <h3>SOS status: <span class="text-success">none <i class="fa fa-check-circle"></i></span></h3>
```
phil-6, line 79: The contrast of this is too low. I would leave this as black and set the card-header to have bg-primary

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+                <div class="card-body text-start">
+                  <p><%= note.content %></p>
+                </div>
+              </div>
+            <% end %>
+          <% end %>
+        </div>
+        <div class="card-footer text-center">
+          <button class="btn btn-primary my-2 px-5" type="button" data-bs-toggle="modal" data-bs-target="#note-modal">
+            Add note
```
IeuanSkinner, line 226: Mixed casing, should be `Add Note`

config/routes.rb
```diff
@@ -16,13 +16,12 @@
...
       root 'dashboard#show', as: :authenticated_user_root
-      get 'home', to: 'users_application#home'
 
-      resources :wba_selves, only: %i[show new create]
-
-      resources :journal_entries, only: %i[index new create] do
-        get 'dashboard', action: :dashboard, on: :collection
+      resources :wba_selves, only: %i[show new create] do
+        resources :wba_self_permissions, only: %i[new create], as: :permissions
+      end
```
IeuanSkinner, line 22: This is wrong, you're reintroducing the wba_self_permissions resources which were stripped and you've also removed the route to home for authenticated users.

app/views/team_members/users/show.html.erb
phil-6, line 95: remove the footer if we're not using it

config/routes.rb
```diff
@@ -16,13 +16,12 @@
...
 
+      resources :journal_entries, only: %i[new create] do
```
IeuanSkinner, line 24: Journal entries index is gone so users won't be able to view their archive and so too is their dashboard.

app/views/team_members/users/show.html.erb
phil-6, line 79: The contrast of this is too low.
Maybe make it black and add bg-success to the header so it ends up like this:
image

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+          <span>Last Sign In</span>
+          <span>
+            <%= @user.last_sign_in_at.present? ? "#{time_ago_in_words(@user.last_sign_in_at)} ago from #{@user_location}" : @user.last_sign_in %>
+          </span>
+        </div>
+      </div>
+    </div>
+
+    <div class="col-12 col-md-6 mb-4">
+      <div class="card h-100 <%= @user.crisis_events.active.empty? ? '' : 'has-crisis' %>">
```
phil-6, line 71: I would consider replacing `has-crisis` with `bg-danger`. This style change (well, just the background colour) should also be applied to the user index.

app/views/team_members/users/show.html.erb
phil-6, line 4: I don't think we need the "User:" prefix here

phil-6 (changes requested):
Looks good overall. Lots of little tweaks though.

app/views/team_members/users/show.html.erb
phil-6: I think `overflow: auto;` will be what you're looking for

app/views/team_members/users/show.html.erb
phil-6: I agree, GitHub had a moment and appeared to lose a lot of comments.

IeuanSkinner (changes requested):
The main thing that cannot be allowed to go through are the changes on `routes.rb`. It looks like you had a merge conflict and set your local changes to be the accepted version.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+        </div>
+        <div class="card-body">
+          <div class="container">
+            <div class="row">
+              <div class="col-12">
+                <div class="container mb-md-0">
+                  <div class="row">
+                    <%= link_to "mailto:#{@user.email}" do %>
+                      <%= @user.full_name %>
+                    <% end %>
```
phil-6, line 23: I don't think we need this as a link

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+                  <div class="row">
+                    <%= link_to "tel:0#{@user.mobile_number}" do %>
+                      <i class="fas fa-phone-alt"></i> <%= "0#{@user.mobile_number}" %>
+                    <% end %>
+                  </div>
+                  <% if @user.release_date.present? %>
+                    <div class="row">
+                      <span><i class="fas fa-calendar"></i> Release: <%= @user.release %></span>
+                    </div>
+                  <% end %>
```
phil-6, line 39: This doesn't need to be wrapped in an if statement. If there is no release date we return "unknown release date" so will look like this which I prefer.
image

app/views/team_members/users/show.html.erb
IeuanSkinner: I've done this on the user index too, personally I like when a webpage has an email address on it I can click the address and it'll open my mail client to send that person an email as that tends to be my objective in clicking.

app/views/team_members/users/show.html.erb
phil-6: I'm not happy with the contrast of the check and times icons here. 
image

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+            View User's Journal
+          </a>
+        </div>
+      </div>
+    </div>
+
+  <!-- End Journal Section -->
+
+    <!-- Notes section -->
+    <!-- Add note modal -->
```
phil-6, line 166: Update the modal so it is 100vw on mobile as there is a lot of wasted space. 
image

Should also add a closing X in the top right to match the SOS modals.

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
+<main class="container team-members users">
```
phil-6, line 1: There is something which is causing the mobile view to be greater than 100% which forces a sideways scroll. I think this is probably needing to set col padding left and right to 0 on small displays
Peek 2021-03-25 09-23

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+      <h2 class="my-3">User: <%= @user.full_name %></h2>
+    </div>
+  </section>
+  <!-- User bio / details section -->
+  <div class="row d-flex justify-content-between">
+    <div class="col-12 col-md-6 mb-4">
+      <div class="card h-100">
+        <div class="card-header">
+            <h3>User info
+            </h3>
```
phil-6, line 13: This could be one-line

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,236 @@
...
+  <div class="row mb-4">
+    <div class="col">
+      <div class="card px-0 d-flex justify-content-center">
+        <div class="card-header">
+          <h3 class="">User wellbeing</h3>
+        </div>
+        <div class="card-body">
+          <div class="row">
+            <div class="col-12 col-lg-5 text-start">
+              <h4>Current wellbeing</h4>
```
phil-6, line 111: I'd like to see the metric names and values here too, like the user view: 
image

maybe add this where the space for the wellbeing history is at the moment and the history can be added to its own section below.

app/views/team_members/users/show.html.erb
phil-6: Yes - but this is wrapping the user's name in the mailto link.

app/controllers/team_members/notes_controller.rb
bmhtech07: I prefer the latter. It feels more useful and the right way of doing it, given we're accessing as a TeamMember. Will resolve this way

app/controllers/team_members/notes_controller.rb
bmhtech07: Echoing Phil, what are the thoughts behind this? Personally, I much prefer UI / UX where the responses are human.

app/controllers/team_members/notes_controller.rb
bmhtech07: Binned it. Think I put it in there to 'force notes' initially

app/javascript/packs/application.js
bmhtech07: Removed Bootstrap and then re-added when trying to get modal working initially (turns out I was missing the 'bs' from the tags)

app/views/team_members/users/show.html.erb
bmhtech07: @phil-6 Is this the flow / functionality that we want? Ie., only link to journal entry, rather than display?

app/views/team_members/users/show.html.erb
bmhtech07: I wonder whether 'check' is the right icon? Maybe '!' if visible and 'lock' (or similar) if not? Think of it from TeamMember's perspective...

app/views/team_members/users/show.html.erb
bmhtech07: Done to keep consistent / symmetrical with adjacent card.

app/views/team_members/users/show.html.erb
bmhtech07: Added `bg-danger` to `card-header`

app/views/team_members/users/show.html.erb
bmhtech07: What to do here? I see no issue in wrapping name in mailto, but happy to follow your leads

app/javascript/packs/application.js
phil-6: In which case, please can you update this file on this branch so its unchanged on the PR.

app/views/team_members/users/show.html.erb
phil-6: Its a good question. 
I think this section would look good if we showed some key stats at the top - about how many journal entries a user has written and how many have been viewed etc. 

Then we could display the most recent journal entry and add a view log for it. This would mean that every time a team member visits a users profile it generates a view log for the users most recent journal entry.

Alternatively we could display a table similar to the journal index page.

I'd appreciate everyone's input on this. What do you think is going to be the most useful to the Include Team Members?
@IeuanSkinner @AlexW555 @bmhtech07

app/views/team_members/users/show.html.erb
phil-6: maybe "eye" and "eye-slash" makes sense for visibility.

app/views/team_members/users/show.html.erb
phil-6: I think you should be able to keep the cards the same height without the footer. The footer section here (color and divider) feels odd if we're not putting anything in it.

app/views/team_members/users/show.html.erb
phil-6: I would prefer the name not to have a link at all. This maintains consistency with the index page. The name links to the profile page. As we're there now, there shouldn't be a hyperlink here. There is also no point in duplicating the link.

app/views/team_members/users/show.html.erb
bmhtech07: I think it would be useful to show the latest entry here, but then of course it's a risk generating a view log as we don't know that it's nesc. been viewed.

Perhaps a truncated version, which encourages click thru to the Entry itself, which is where the view log is generated?

app/views/team_members/users/show.html.erb
bmhtech07: I agree, but I think an element of "warning, the User can see this". Perhaps 'eye' `text-danger` and 'eye-slash'

app/views/team_members/users/show.html.erb
bmhtech07: Yes, that's what the h-100 is doing. I just mean visually, it looks odd without a footer when the cards are equal heights.

image

app/controllers/team_members/notes_controller.rb
IeuanSkinner: It's software not a person, you're lying to people and patronizing them when you say "we" can't create a note because there is no "we".

`Note could not be created.` is factually true.

app/views/team_members/users/show.html.erb
phil-6: I prefer this. I think its weirder to have the card there with the empty footer.

@AlexW555 @IeuanSkinner  what do you think?

app/views/team_members/users/show.html.erb
IeuanSkinner: A team member should be able to see how many **total** journal entries a user has and how many **unread** journal entries they have and they should be able to click through to an index of all those entries `/users/:user_id/journal_entries` and `/users/:user_id/journal_entries/unread`, these views are just the current journal entries index view filtered and all of this needs to adhere to the permissions structure.

As for what to display on the users member page showing the latest unread entry in terms of time, feeling (emoji) and viewed status is fine but you can't show the text without registering a view log and you can't consider it read if they haven't expressly opened it even a 'truncated' version so don't show that. A link to 'Read journal entry' along with the aforementioned info should be fine in that section.

app/views/team_members/users/show.html.erb
IeuanSkinner: Oh I completely missed that this was wrapping the users name at the top of the card. Yeah this shouldn't be a link once you're at the users member view.

app/views/team_members/users/show.html.erb
bmhtech07: Why can't you show the truncated text from the entry without registering a view log?

I agree, it's not right to register a view log unless they've expressly clicked through to read.

app/controllers/team_members/notes_controller.rb
phil-6: We're going down the UX rabbit hole here! (And I love it)

The answers on [this stack exchange post](https://ux.stackexchange.com/questions/31359/should-error-messages-apologize) are interesting, although not 100% related to this.

I still hold [Neisen](https://www.nngroup.com/articles/error-message-guidelines/) as the gold standard here, but again not giving an exact answer to this.

Doing some reading - the general opinions seems to be that you absolutely should humanise your interfaces and error messages, but nowhere I can see from an initial reading goes into using "we" so maybe something like:
"Unfortunately the your note could not be created. Please try again. Here's the content you tried to submit if you want copy it to try again"

app/views/team_members/users/show.html.erb
phil-6: > Why can't you show the truncated text from the entry without registering a view log?
Because, if they've read some of it they have had access to potentially sensitive information and that should be logged.

phil-6: For this page I agree with Ieuan.
Some key info at the top "User has written X journal entries, X of which you haven't read. X have been written in the last month. Here are the most recent 5 entries" Then a list of the most recent 5 which emoji, timestamp and viewed status. Then a button to view all entries at the bottom of the card.

IeuanSkinner: There's no amount of truncation you can do that won't potentially have meaningful data in the entry, for example if someone just writes "want 2 die" truncation wouldn't prevent that appearing and so the platform needs to know if a team member has potentially seen that in case the user comes to Include and says "team member has been exploiting my suicidal thoughts".

If the team member can claim "I wasn't even aware of such a thing" and the platform corroborates that due to no view log that's bad, the simplest way to avoid this is the principle of "do not show any of the entry unless the team member has explicitly clicked to view it".

app/controllers/team_members/notes_controller.rb
AlexW555: I would agree with Ieuan that using "we" is very patronizing if I had that as an error message. I would have something like Ieauns with maybe some hint in phils to copy the content incase if it's likely they will lose it

app/views/team_members/users/show.html.erb
IeuanSkinner: I agree the above actually looks fine to me, if there were a blank footer section on the SOS status card I'd be wondering what information _should_ be appearing there but isn't.

AlexW555: I don't really find the bottom bit hugely necessary if you cant think of anything to be there. Perhaps something could go there in the future, if we want to link to close/expand a crisis etc. But for now, I would be happy just leaving it without a footer 👍🏻

app/views/team_members/users/show.html.erb
bmhtech07: I see. I was thinking you meant from a technical perspective. I fully understand.}

Will adjust per comments.

bmhtech07:
All changes per above.

image

phil-6 (approved):
Overall, this looks good to me and I'm happy for it to be merged to main.

app/assets/stylesheets/global.scss
```diff
@@ -18,8 +21,15 @@ body#overboot {
     font-weight: 800;
     text-transform: lowercase;
   }
-  background-color: $black;
-  font-family: $font;
+
+  .row {
+    --bs-gutter-x: 0;
+    --bs-gutter-y: 0;
+    @media (min-width: 576px) {
```
IeuanSkinner, line 28: We have a variable that can be used for this `$screen-sm`

app/controllers/team_members/users_controller.rb
```diff
@@ -1,15 +1,16 @@
 module TeamMembers
   # app/controllers/team_members/users_controller.rb
   class UsersController < PaginationController
-    before_action :user, except: :index
+    before_action :user, except: %i[index]
```
IeuanSkinner, line 4: If it's a single instance `%i[...]` is not needed.

app/controllers/team_members/users_controller.rb
```diff
@@ -58,6 +59,50 @@ def resources
...
+    def wba
+      @wba = @user.last_wellbeing_assessment
+    rescue ActiveRecord::RecordNotFound
+      session notice: 'No wellbeing assessment could be found'
+    end
+
+    def wellbeing_metrics
+      if @wba.present?
+        (@wellbeing_metrics = @wba.wba_scores.includes(:wellbeing_metric))
+      end
```
IeuanSkinner, line 91: Guard clause should be used here and is flagged by RuboCop

```
def wellbeing_metrics
  return unless @wba.present?
  
  @wellbeing_metrics = @wba.wba_scores.includes(:wellbeing_metric)
end
```

app/javascript/packs/line_chart.js
```diff
@@ -0,0 +1,47 @@
...
+            tooltips: {
+                callbacks: {
+                    label: function (tooltipItem, data) {
+                        return `${data.labels[tooltipItem.index]}: ${scale[tooltipItem.yLabel - 1].description}`
+                    }
+                }
+            }
+        }
+    })
+}
```
IeuanSkinner, line 47: This can be deleted as it's unused and is just a duplication of the `polar_area_chart.js`

app/controllers/team_members/notes_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    # PATCH/PUT /notes/1
+    def update
+      redirect_to user_path(@user), notice: @note.update!(note_params) ? 'Note updated!' : "The note couldn't 'be added. Please try again."
+    end
+
+    # DELETE /notes/1
+    def destroy
+      @note.destroy
+      redirect_to notes_url, notice: 'Note was successfully destroyed.'
```
AlexW555, line 24: Maybe use "Note successfully deleted/removed" But destroyed does sound pretty cool 😂

AlexW555 (commented):
LGTM ✅

config/routes.rb
```diff
@@ -52,6 +52,10 @@
         put 'unpin', action: 'unpin', on: :member, as: :unpin
       end
 
+      resources :users, only: :show, as: :user do
+        resources :notes, only: :create, as: :add_note
+      end
```
IeuanSkinner, line 57: There's 2 `:users` resources being declared in this block now, the nested resource `:notes` can be place in the block above.

app/controllers/team_members/notes_controller.rb
IeuanSkinner: I've only just spotted this, currently you're not fetching the notes from the `current_team_member` meaning any team member can delete any note and any team member can update any note, we need it so a team meber can only edit and delete their own notes. Even then we should probably check whether we really want *true* deletion or to just mark the record as deleted but keep its contents around.

app/controllers/team_members/users_controller.rb
```diff
@@ -58,6 +59,50 @@ def resources
...
+    def user
+      @user = User.includes(:notes).find(params[:id])
+    end
+
+    def user_count
+      @user_count = User.count
+    end
+
+    def user_notes
+      @user_notes = @user.notes.order(created_at: :desc)
```
IeuanSkinner, line 75: Here you're pulling through the notes and in the view you access the team member associated with that note so they can be added to the card heading but because you haven't included the team member this leads to a DB read for every single note.

You fix this by including the team member here:
```
def user_notes
  @user_notes = @user.notes.includes(:team_member).order(created_at: :desc)
end
```

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,258 @@
...
+                  </div>
+                <% end %>
+              </div>
+            </div>
+          </div>
+        </div>
+        <div class="card-footer d-flex justify-content-between">
+          <span>Last Sign In</span>
+          <span>
+            <%= @user.last_sign_in_at.present? ? "#{time_ago_in_words(@user.last_sign_in_at)} ago from #{@user_location}" : @user.last_sign_in %>
```
IeuanSkinner, line 60: This just says `from XX` for me?

Screenshot 2021-03-31 at 15 27 49

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,258 @@
...
+          </span>
+        </div>
+      </div>
+    </div>
+    <div class="col-12 col-md-6 mb-4">
+      <div class="card h-100 <%= @user.crisis_events.active.empty? ? '' : 'has-crisis' %>">
```
IeuanSkinner, line 66: This is the same thing as the line below `@active_crisis.empty?` except it requires an additional DB read.

app/controllers/team_members/users_controller.rb
```diff
@@ -58,6 +59,50 @@ def resources
...
+    end
+
+    def user_location
+      @user_location = Timeout::timeout(5) { Net::HTTP.get_response(URI.parse('http://api.hostip.info/country.php?ip=' + @user.last_sign_in_ip )).body } rescue "Unknown"
+    end
```
phil-6, line 80: I think this can be improved as follows: 
```ruby
def user_location
  return @location = { 'city' => 'unknown' } unless @user.last_sign_in_ip.present?
  @location = Timeout.timeout(5) { JSON.parse(Net::HTTP.get_response(URI("http://ip-api.com/json/#{@user.last_sign_in_ip}" )).body) }
  # @location["city"]
  # @location["zip"]
end
```

app/views/team_members/users/show.html.erb
```diff
@@ -0,0 +1,258 @@
...
+                </tbody>
+              </table>
+            </div>
+          <% else %>
+            <span>The User hasn't completed any Journal entries yet </span>
+          <% end %>
+        </div>
+        <div class="card-footer text-center">
+          <button class="btn btn-primary my-2 py-2 px-5" type="button">
+            <%= link_to "/journal_entries?query=#{@user.first_name}+#{@user.last_name}" do %>
```
IeuanSkinner, line 185: This string templating isn't needed

`journal_entries_path(query: @user.full_name)`

app/controllers/team_members/notes_controller.rb
bmhtech07: That's correct, but there are currently no routes / UI to allow this. This was standard CRUD functionality that I believe came as standard on the controller.

I will remove for now, and then spin in to a new ticket.

app/controllers/team_members/users_controller.rb
```diff
@@ -58,6 +59,52 @@ def resources
...
+
+    def user_notes
+      @user_notes = @user.notes.includes(:team_member).order(created_at: :desc)
+    end
+
+    def user_location
+      return @location = { 'city' => 'unknown' } unless @user.current_sign_in_ip.present? && ['127.0.0.1', '::1'].exclude?(@user.current_sign_in_ip)
+
+      @location = Timeout.timeout(5) { JSON.parse(Net::HTTP.get_response(URI("http://ip-api.com/json/#{@user.current_sign_in_ip}" )).body) }
+    end
```
phil-6, line 82: Been thinking about this one @bmhtech07, it still hurts my head too much so I think it could be extracted / improved to this: 

'''ruby
    def user_location
      return @location = { 'city' => 'unknown' } if invalid_ip

      @location = Timeout.timeout(5) { JSON.parse(Net::HTTP.get_response(URI("http://ip-api.com/json/#{@user.current_sign_in_ip}" )).body) } rescue { 'city' => 'unknown' }
    end

    def invalid_ip
        return true  if @user.current_sign_in_ip.blank? 

         ['127.0.0.1', '::1'].include?(@user.current_sign_in_ip)
    end
'''

config/routes.rb
```diff
@@ -50,8 +49,13 @@
         put 'increment', action: 'increment', on: :member, as: :increment
         put 'decrement', action: 'decrement', on: :member, as: :decrement
         put 'unpin', action: 'unpin', on: :member, as: :unpin
+        resources :notes, only: :create, as: :add_note
       end
 
+      # resources :users, only: :show, as: :user do
+      #   resources :notes, only: :create, as: :add_note
+      # end
```
IeuanSkinner, line 57: Can be deleted.

IeuanSkinner (approved):

PR #43, opened 2021-03-26
IJ-114 Added search results display to index pages.

IeuanSkinner:
All pages that are searchable will now display text outlining the results of that search.

Screenshot 2021-03-26 at 08 45 49


phil-6 (approved):
Looks Great!

PR #44, opened 2021-03-26
IJ-115 Added limit controls to index pages.

IeuanSkinner:
I've added a limit select field that allows a user to select the total number of resources they want to display in an index.

Screenshot 2021-03-26 at 09 53 11


phil-6 (changes requested):
Code looks good to me but still a few bugs.

Controls also show up on User Journal Archive but don't work (and the numbers don't make sense) 
image

Wellbeing Assessment Index controls don't work correctly
image

I also think it might make sense to have the controls as 6 - 12 - 30 - 60 so it matches the standard 2x column layout better. In order to get it working correctly for the User Journal Page maybe have it has `default-limit = x` then control doing `x - 2x - 5x - 10x`

phil-6:
Page size is not preserved when moving between pages.

Peek 2021-03-29 10-15

IeuanSkinner:
> Page size is not preserved when moving between pages.

Yeah I spotted this and committed a fix, if you pull the latest for this branch and try again 👍

app/views/team_members/crisis_events/index.html.erb
```diff
@@ -6,6 +6,7 @@
     <div class="col">
       <h3>Closed Crisis Events</h3>
       <%= render 'shared/search_results', query: params[:query], count: @count %>
+      <%= render 'shared/limit', multiple: 5 %>
```
phil-6, line 9: this now feels like its being defined in both the front and back end. Opens up the possibility of a mis match between controls and backend. If we move all responsibility for how many records to display to the front end it eliminates this risk.

phil-6 (commented):
I'm happy with this, it functions really well. But I think we can improve it and we might be mixing concerns slightly.

app/views/team_members/crisis_events/index.html.erb
IeuanSkinner: Yeah I don't know why I did this, it's now singularly defined in the controller for the view.

AlexW555 (commented):
Really Impressive, LGTM ✅

I only have two small design changes which aren't at all huge just things I noticed. 
1. When there is only a single page, the "1" is slightly off-center.
Screenshot 2021-03-31 at 11 27 27 am
2. When you navigate to pages 2,3,4. The dropdown resets to the default value. 

https://user-images.githubusercontent.com/17786854/113131141-bcd4dc80-9214-11eb-9836-0d2749f1eac4.mov

IeuanSkinner:
>     2. When you navigate to pages 2,3,4. The dropdown resets to the default value.

@AlexW555 I'm also not seeing this, can you make sure you've pulled the latest version of this branch and try to replicate.

phil-6 (approved):

PR #45, opened 2021-03-29
IJ-56 team member create wba + IJ-138 update user view to use wellbeing chart partial

IeuanSkinner:
When a team member clicks on a WBA they'll now be taken to that WBA's member view, if the WBA was completed by a team member that info is now displayed in the WBA member view. You can click on the 'Add New WBA' button in the WBA member view and be taken to a view to fill out a new WBA for the given user.

Screenshot 2021-03-29 at 11 19 50

Screenshot 2021-03-29 at 11 20 02

This PR has also seen a sizable refactor of the WBA charts to use a partial rather than duplicating markup all over.


phil-6:
There is still the bug that the wba graphs don't render when traveling "back" to the page. @IeuanSkinner did you ever get to the bottom of what was causing this?

phil-6:
After merging main -> this branch I'm now getting this error: 
image

IeuanSkinner:
@phil-6 yeah I'm working on that at the moment IJ-138

IeuanSkinner:
@phil-6 should be fixed if you pull latest change

IeuanSkinner:
> There is still the bug that the wba graphs don't render when traveling "back" to the page. @IeuanSkinner did you ever get to the bottom of what was causing this?

If you navgiate with the in-platform buttons I don't think this happens whereas with the browsers back button it does as the browser caches the markup so the JS code recognises the canvas as having been rendered so doesn't rerender it. You can remove this check and that fixed the back button issue but then you get an issue where the charts render twice when you hit the index page.

Bit of a headache enducer but it can be solved but I think that should be considered a seperate bit of work.

Removing this check

phil-6:
Is there a limit as to how often a team member can create a WBA? If not, please can we add something similar to the user's limit.

Please can we add a link from the show wba to show user. Maybe a "view profile" button

> Bit of a headache enducer but it can be solved but I think that should be considered a seperate bit of work.

Please can you create a ticket for this in Jira

phil-6 (commented):
Looks great overall, few changes in previous comment

IeuanSkinner:
> Is there a limit as to how often a team member can create a WBA? If not, please can we add something similar to the user's limit.

There isn't but I've put in place the same check as exists for the users. I've also clocked that the way the check works means only 1 WBA can be entered per user per day for **both** team members and users meaning if a user enters 1 for themself on a day then a team member can't also put one in and vice versa, personally I think this actually makes sense as there's no good reason why a users wellbeing on a given day should be recorded by both themselves _and_ a team member.

> Please can we add a link from the show wba to show user. Maybe a "view profile" button

Done,  this will appear in the top-right if a new assessment can't be filled out otherwise it's in the bottom right.

> > Bit of a headache enducer but it can be solved but I think that should be considered a seperate bit of work.
> 
> Please can you create a ticket for this in Jira

Done IJ-140

phil-6:
> There isn't but I've put in place the same check as exists for the users. I've also clocked that the way the check works means only 1 WBA can be entered per user per day for **both** team members and users meaning if a user enters 1 for themself on a day then a team member can't also put one in and vice versa, personally I think this actually makes sense as there's no good reason why a users wellbeing on a given day should be recorded by both themselves _and_ a team member.

yeah, that makes sense. Thanks1

> Done, this will appear in the top-right if a new assessment can't be filled out otherwise it's in the bottom right.

> Done IJ-140

Thank you

phil-6 (approved):

PR #46, opened 2021-04-01
IJ-135 add external font import to get rubik working on chromium base…

phil-6:
…d browsers

really not sure why this is needed, but it works now ¯\_(ツ)_/¯


IeuanSkinner (approved):
LGTM, I checked too see if there's anywhere we use the Rubik font with a different font-weight and we don't appear too so this should suffice.

bmhtech07:
LGTM

PR #47, opened 2021-04-01
Ij 137 replace last sign in

bmhtech07:
Changed user cards to:

- use current_sign_in (rather than last_sign_in) and set the falsey condition to 'never'.
- Header background to $danger in .has-crisis class per conversation with Phil
- Strip card border.

Reflected changes in team_members/user/show view.


app/views/team_members/users/show.html.erb
```diff
@@ -66,7 +66,8 @@
       <% if @active_crisis.empty? %>
         <div class="card h-100 border-0">
           <div class="card-header bg-success">
-            <h3>SOS status: none <i class="fa fa-check-circle"></i></h3>
+            <h3>SOS status:
+              <span class="h3">none <i class="fa fa-check-circle"></i></h3>
```
IeuanSkinner, line 70: Missing closing span tag

IeuanSkinner:
@phil-6 should we use this ticket as an opportunity to convert all places we've used `last_sign_in_at` to `current_sign_in_at`?

phil-6:
> @phil-6 should we use this ticket as an opportunity to convert all places we've used `last_sign_in_at` to `current_sign_in_at`?

Yes, should it be appropriate which I think in most cases will be. @bmhtech07 are you okay to do the extra refactoring here?

phil-6 (changes requested):
As per the comment above, @bmhtech07 please can you take ownership of the refactor for the `sign_in_at` variables

bmhtech07:
> As per the comment above, @bmhtech07 please can you take ownership of the refactor for the `sign_in_at` variables

Sorry, I missed this previously.

Yes, no problem.

bmhtech07:
Refactoring done on this.

I believe there was only one other instance (on the User model) where `current` was more appropriate than `last`.

phil-6 (approved):

IeuanSkinner (approved):

PR #49, opened 2021-04-07
IJ-97 + IJ-91 

IeuanSkinner:
I've added in the ability for admins to pause team member accounts via a button on the team members index. This button has a confirmation warning explaining that the team member will be locked out of the system whilst paused and the team member is informed when they attempt to access the system why that can't.

I've also gone ahead and removed the box-shadow properties from all buttons whilst here which is IJ-91.

Screenshot 2021-04-07 at 10 42 57

Screenshot 2021-04-07 at 10 43 14

Screenshot 2021-04-07 at 10 43 54

### DIFF
--- a/app/controllers/team_members/team_members_controller.rb
+++ b/app/controllers/team_members/team_members_controller.rb
@@ -1,11 +1,13 @@
 module TeamMembers
   # app/controllers/team_members/team_members_controller.rb
   class TeamMembersController < AdminApplicationController
-    before_action :team_member, only: %i[show approve_team_member toggle_admin]
-    before_action :unapproved_team_members, :approved_team_members, only: :index
+    before_action :team_member, except: :index
 
     # GET /team_members
     def index
+      @unapproved_team_members = TeamMember.unapproved.order(created_at: :desc)
+      @approved_team_members = TeamMember.approved.order(created_at: :desc)
+
       render 'index'
     end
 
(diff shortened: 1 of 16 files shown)


IeuanSkinner:
There's a bug in here with the flash messages not appearing that has been fixed in the IJ-98 branch.

app/controllers/team_members/team_members_controller.rb
```diff
@@ -1,11 +1,13 @@
...
   class TeamMembersController < AdminApplicationController
-    before_action :team_member, only: %i[show approve_team_member toggle_admin]
-    before_action :unapproved_team_members, :approved_team_members, only: :index
+    before_action :team_member, except: :index
 
     # GET /team_members
     def index
+      @unapproved_team_members = TeamMember.unapproved.order(created_at: :desc)
+      @approved_team_members = TeamMember.approved.order(created_at: :desc)
+
```
phil-6, line 10: Extra empty line

phil-6 (approved):
Overall LGTM.
Only question is around semantics paused / unpaused or active / suspended

app/controllers/team_members/team_members_controller.rb
IeuanSkinner: This is intentional, an extra space between the instance variable assignments and controller render action can improve readability.

IeuanSkinner:
> Only question is around semantics paused / unpaused or active / suspended

The Jira ticket specifically says pause/paused.

PR #50, opened 2021-04-07
IJ-98 - Team Member Rejection

IeuanSkinner:
Admin team members can now also reject team member sign ups which will destroy that team member record. I've also fixed the flash messages not appearing if they're of any type other than notice.

Screenshot 2021-04-07 at 12 07 41

### DIFF
--- a/app/controllers/team_members/team_members_controller.rb
+++ b/app/controllers/team_members/team_members_controller.rb
@@ -25,12 +25,10 @@ module TeamMembers
     def reject
       return if @team_member.approved
 
-      name = @team_member.full_name
-
       if @team_member.destroy
-        redirect_to team_members_path, flash: { success: "#{name} has been rejected" }
+        redirect_to team_members_path, flash: { success: "#{@team_member.full_name} has been rejected" }
       else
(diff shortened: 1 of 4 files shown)


app/controllers/team_members/team_members_controller.rb
```diff
@@ -21,6 +21,19 @@ def approve
       @team_member.update(approved: true) ? success(true, 'approved') : failure('approval')
     end
 
+    # PUT /team_members/:id/reject
+    def reject
+      return if @team_member.approved
+
+      name = @team_member.full_name
```
phil-6, line 28: do we need to assign this before the conditional?

phil-6 (approved):

app/controllers/team_members/team_members_controller.rb
IeuanSkinner: Destroying the record doesn't appear to clear the object immediately so we can still call `full_name` in the body of the conditional so no we don't. To me it makes more sense to take a copy of this value before destroying the attached record so you know you can reference it post-destroy but the rails internals seem to operate differently.

PR #51, opened 2021-04-07
Refactor to reduce use of action filters

phil-6:
Hi All,
Putting this up for PR early as its not actually work related to this branch! This is a refactor of the team_members/pagination_controller and the files which extend it.

The aim here is to increase the overall readability of the code to increase the overall maintainability. Please let me know if you think if this achieves this - and also see if my changes have broken anything!
I think I've manually tested all the functionality but it is likely I've missed something or made a last minute change and not fully retested!

Questions and comments encouraged!


IeuanSkinner:
This breaks the Journal Entry View Logs index and also the Crisis Events index as you're relying on the classes that extend the pagination controller to call the `count` and `limit_resources` functions at the end of their `resources` function which you haven't then implemented for these controllers, this also breaks the abstraction as count and limit resources are supposed to be universal for paginated indexes so any logic relating to them shouldn't exist at the subclass level.

Overall I don't think this refactor is a good one as it doesn't improve readability, it's much easier to read:

`:this_should_happen, :then_this, :then_this ... before ... :this_action`

than it is to hop around the functions and figure out who is in charge of calling what i.e. why does `last_page` get called by `limit_resources`? Why is `limit` being called in multiple places when it should only require 1 operation to set the limit from the query param?

phil-6:
Thanks for spotting the View Logs and Crisis Events bugs - they will need fixing before this gets merged. 

Why should count and limit resources be universal? From the way I understand it, they look at the resources which have just been set, count the total and then limit it down to the number which should be rendered in the index. To me that's directly related to the `@resources` variable and as such should be close to them in the code. 
They could be extracted back to the pagination controller above the "render or redirect" line - presuming we stick with this approach.

More generally, I still strongly disagree that it is easier to read with the filters in place. Previously we had this as the filters before the index action on the wellbeing assessment controller: 

```ruby
    before_action :query_params, :page, :query, :limit, :offset, :team_member, :wellbeing_assessments, only: :index

    before_action :resources, only: :index, unless: -> { @query.present? }
    before_action :search, only: :index, if: -> { @query.present? }

    before_action :count, :last_page, :limit_resources, :redirect, only: :index
```

and this in the parent pagination controller:
```ruby
    before_action :query_params, :page, :query, :multiple, :limit, :offset, :resources,
                  :count, :last_page, :limit_resources, :redirect, only: :index
```

This was the key section of code I was unhappy with. It's not clear to me what variables are being set and a lot of the filters are intrinsically linked.

One of the main issues I encountered is that I wanted to add the `:user` filter after ':team member` and before `wellbeing_assessments` and everything else gets called. However, we also have the line `before_action :user, :wellbeing_metrics, only: %i[new create]` which overwrites the `:user` call when its used as an index action filter. With the filters in their original state I am unaware of a way to get the order of operations to run correctly.

What I have tried to do with the refactor is group intrinsically linked method calls together so, although if you need to delve into every section of the code you might do a little bit more "hopping around" but you should be able to see what the index method is doing in a more glance-able way.

The two sections above have changed to:

```ruby
def index
    team_member
    user
    query_params[:query].present? ? search : resources
    super
end
```

Which, to me, clearly says, get team member, get user, if there's a query then search, else get resources, then do the parent action:

```ruby
 def index
    @page = page
    @query = query_params[:query]
    multiple unless @multiple
    resources unless @resources
    @resources.present? ? render('index') : redirect
end
```
to me this is clear: set page variable, set query variable, if multiple hasn't already been set, then set it, if resources haven't already been set then do that method, then if there are resources render index, else redirect. 

You might be right about the limit resource, I'll see if there is any performance difference between these approaches. Same goes for the `last_page` call, that's probably not in the correct place, it might be better at the end of the `limit` function. I'll take another look at those.

Overall, I still am much happier with this general approach. 
Does what I've explained above sway your position at all @IeuanSkinner ?

@AlexW555, @garethdandrews, @bmhtech07, do you have any thoughts on this? What do you feel is easier to read and understand?

IeuanSkinner:
So for anything that's paginated theres going to be some kind of `resource` that is getting displayed within a page hence why we have those `resources` set by the subclass as that class knows what the `resource` is. The way you count those resources once instantiated is always the same it's always `@count = @resources.count` and the way you limit those resources is again always the same it's `@resources = @resources.offset(@offset).limit(@limit)` hence what I mean by "universal" as it doesn't matter _what_ the resource is, the operation is always the same.

What your refactor is doing is requiring every subclass to now have the 2 lines 
```
    @count = count
    limit_resources
```
at the end of the resources function which means every subclass is going to duplicate these lines.

If I'm understanding correctly you want a route of `/users/:user_id/wellbeing_assessments` like we have for team members which provides the index view of wellbeing assessments filtered by user? The confusion being that user is currently set in one before_action filter for new and create actions but now we want it happening before the index too which has these chained actions as some actions are dependant on each other (limit depends on multiple, offset depends on limit etc.). 

This is really simple as the user is constructed from query params so isn't dependent on any code having evaluated so you only need to extract the user setup to before the other actions. I've done this on the branch feature/IJ-139-test, if you run this branch you can go to the aforementioned route.

IeuanSkinner:
Using this as a guide I've put together a refactor [here](https://github.com/Legal-Innovation-Lab-Wales/include-journey/pull/52).

PR #52, opened 2021-04-08
Feature/ij 139 refactor

IeuanSkinner:
I've refactored the pagination classes in the team members module so that we're only using `before_action` filters where an action needs to occur before mutliple differing actions.

I've also fixed a bug with the team member new wellbeing assessments here.


phil-6 (approved):
I like this lots. If you merge this in main, I'll update my branch and crack on with the rest of IJ-139.

Thanks Ieuan, great work!

PR #53, opened 2021-04-08
IJ-87 User Wellbeing History Chart

IeuanSkinner:
I've added a new segment to the team member user view which displays a history of the users wellbeing assessments as a line chart with the time series as the x-axis, the score given to the metric as the y-axis and the metric itself represented as a line on this chart.

The team member can view different metrics by toggling them as hidden/shown in the chart legend (the first metric loaded displays on initialisation with the remainder hidden by default).

As our data is randomly generated from the seed the lines we end up with tend to erratically jump across the values, this may end up being representative of how users interpret their wellbeing but my gut feeling is that we're more likely to see gradual changes over time with real data.

Screenshot 2021-04-08 at 15 36 35

### DIFF
--- a/app/javascript/packs/wellbeing_history_chart.js
+++ b/app/javascript/packs/wellbeing_history_chart.js
@@ -49,8 +49,8 @@ fetch(`${location}/wba_history`, {
                         }
                     },
                     ticks: {
-                        min: json.labels[0],
-                        max: json.labels[json.labels.length - 1]
+                        bounds: 'ticks',
+                        source: 'auto'
                     },
                     gridLines: {
                         zeroLineColor: 'gray',
--- a/app/javascript/packs/wellbeing_history_chart.js
+++ b/app/javascript/packs/wellbeing_history_chart.js
@@ -1,0 +1,36 @@
+fetch(`${location}/wba_history`, {
+    headers: {
+        'Content-Type': 'application/json'
+    }
+})
+    .then(data => data.json())
+    .then(json => {
+        const schemeCategory10 = [
+            '#1f77b4', '#ff7f0e', '#2ca02c', '#d62728', '#9467bd',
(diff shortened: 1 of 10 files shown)


app/javascript/packs/wellbeing_history_chart.js
```diff
@@ -0,0 +1,85 @@
...
+    { description: "Dreadful", colour: "#E64D52" },
+    { description: "Rubbish", colour: "#E86754" },
+    { description: "Bad", colour: "#EC8754" },
+    { description: "Mediocre", colour: "#F0A656" },
+    { description: "Fine", colour: "#DFC54C" },
+    { description: "Good", colour: "#BFCA48" },
+    { description: "Great", colour: "#9ECB46" },
+    { description: "Superb", colour: "#BFCB43" },
+    { description: "Perfect", colour: "#5DAD3A" }
+]
```
phil-6, line 12: as this is now used in multiple places - should we extract it out into the database?

db/seeds.rb
```diff
@@ -203,7 +205,9 @@
         WbaScore.create!(
           wellbeing_assessment: wellbeing_assessment,
           wellbeing_metric: wellbeing_metric,
-          value: rand(1..10)
+          value: rand(1..10),
+          created_at: Date.today - (wba_count - 1),
+          updated_at: Date.today - (wba_count - 1)
```
phil-6, line 210: you could pluck this commit and use it for the seed as it introduces some variability into the created at dates.
https://github.com/Legal-Innovation-Lab-Wales/include-journey/pull/54/commits/931361065653b8a151079eeec98eadd6d0afb3de

Do we need to set the updated_at?

phil-6:
This gets very noisy with a lot of wellbeing assessments.

image

IeuanSkinner:
This is in the process of being reworked significantly.

PR #54, opened 2021-04-09
Feature/IJ-139 user wellbeing index

phil-6:
image

A different view for the user's wellbeing index. 
Not designed to be used for analysis over a user's wellbeing but to allow for easy navigation to a specific wellbeing assessment. The aim was to make it easier to access historical wellbeing assessments. There are likely to be other ways to navigate to these in the future. Each row in the table links to the specific assessment and there a tooltips over each of the colour squares.

I'm putting this up for PR so you can have a look at how it works in person instead of just looking at screen shots. I quite like it because it IS different to the card view, although definitely not as useful for analytics or anything but maybe a useful navigation tool. 

----

Updated to add in the card view.

image

The card view is now the default. The grid / table view still exists and can be accessed via a hidden button. If include don't like this I'll kill it but I'd like to leave it there until we demo it to Include.


phil-6:
View is now cards as per discussion.
image

app/views/team_members/wellbeing_assessments/index.html.erb
```diff
@@ -135,6 +164,21 @@
...
+            document.getElementById("card-view").classList.add('d-none');
+            document.getElementById("table-view").classList.remove('d-none');
+        } else if (document.getElementById("card-view").classList.contains('d-none')) {
+            document.getElementById("table-view").classList.add('d-none');
+            document.getElementById("card-view").classList.remove('d-none');
+        } else {
+            alert("Somethings gone wrong with the View, Sorry!")
+        }
+    }
+</script>
```
IeuanSkinner, line 184: Not a fan of having this hidden mode in the view 🤨 

```
<script>
    function toggleViewMode() {
        document.getElementById("table-view").classList.toggle('d-none')
        document.getElementById("card-view").classList.toggle('d-none')
    }
</script>
```

phil-6: did not know there was a `classList.toggle` function, thanks!

As much as I love super secret hidden buttons, I do see the point about not having hidden views - its not needed. 
That being said, having met Mel I think she is going to love the fact there is a hidden button so I vote to leave it in until we demo it to them and see what she says.

app/controllers/team_members/wellbeing_assessments_controller.rb
```diff
@@ -119,5 +119,14 @@ def wellbeing_assessment_today
...
       @wellbeing_metrics = WellbeingMetric.all.order(:created_at)
     end
+
+    def subheading_stats
+      @count_in_last_week = @resources.where('created_at >= ?', 1.week.ago).size
+      @count_in_last_month = @resources.where('created_at >= ?', 1.month.ago).size
+      return unless @user
+
+      @count_by_team_member = @resources.count { |tm| tm.team_member_id.present? }
+      @count_by_user = @resources.count { |tm| tm.team_member_id.nil? }
```
IeuanSkinner, line 129: The block param name implies the object is a type of team member but it's a wellbeing assessment isn't it?

```
@count_by_team_member = @resources.count { |wba| wba.team_member_id.present? }
@count_by_user = @resources.count { |wba| wba.team_member_id.nil? }
```

app/views/team_members/wellbeing_assessments/index.html.erb
```diff
@@ -10,68 +10,168 @@
...
+                                                                       when 7
+                                                                         'Good'
+                                                                       when 8
+                                                                         'Great'
+                                                                       when 9
+                                                                         'Superb'
+                                                                       when 10
+                                                                         'Perfect'
+                                                                       else
+                                                                         'Unknown'
```
IeuanSkinner, line 90: Can we instantiate this as a hash on the back-end that gets reused in the view to map these instead?

```
@wba_values = {1 => "Abysmal", 2 => "Dreadful", 3 => "Rubbish", ... }

<%= @wba_values[score.value] %>
```

IeuanSkinner (approved):
LGTM

PR #55, opened 2021-04-09
IJ-143 Pagination Refactor

IeuanSkinner:
I've refactored the pagination so that this functionality is integrated into a controller via composition rather than inheritance which has cleaned up a lot of the repeating code.


app/controllers/concerns/pagination.rb
```diff
@@ -0,0 +1,68 @@
+# app/controllers/concerns/pagination.rb
+module Pagination
+  extend ActiveSupport::Concern
+  included do
+    before_action :pagination, only: :index
+  end
+
+  protected
+
+  # rubocop:disable Metrics/AbcSize
```
phil-6, line 10: I love this, sometimes rubocop does need to just do one.

app/controllers/team_members/users_controller.rb
```diff
@@ -1,17 +1,15 @@
...
 
     # GET /users
-    def index
-      @resources_per_page = 6
-      @pinned_users = current_team_member.pinned_users.order(:order)
-      @active_users = User.where(last_sign_in_at: (Time.zone.now - 30.days)..Time.zone.now).count
-      @user_count = User.count
-      super
-    end
+    def index; end
```
phil-6, line 12: do we need this?

app/controllers/team_members/wellbeing_assessments_controller.rb
```diff
@@ -16,18 +12,21 @@ class WellbeingAssessmentsController < PaginationController
     # GET /wellbeing_assessments
     # GET /team_members/:team_member_id/wellbeing_assessments
     # GET /users/:user_id/wellbeing_assessments
-    def index
-      @resources_per_page = 6
-      super
-    end
+    def index; end
```
phil-6, line 15: I don't think the empty methods do anything

phil-6 (approved):
This is a really great refactor and a great piece of work to get done is such a short amount of time. Tons of code duplication removed and overall a really great implementation for pagination.
Great work Ieuan!

My only request is that this gets merged after the other open PRs as there are going to be some conflicts!

app/controllers/team_members/users_controller.rb
IeuanSkinner: They're not needed but they're referenced by the action filters so your IDE will mark them unless you suppress that inspection,  do you want them left in or removed given this?

phil-6: ahh, I see, lets leave them in to keep JetBrains happy. :airplane: :brain: :smile:

PR #57, opened 2021-04-16
Hotfix/IJ-151 update user bug

phil-6:
## Problem

Previously attempting to update a user or team member account would send the request with a HTML DELETE method instead of a PUT method.

This could have led to the inadvertent deletion of accounts.

## Cause

This was caused due to the `button_to` for deleting being inside the the `form_for` update, so the submit button was pulling the html method from the last element which was delete.

## Fix

I've moved the `button_to` to outside the `update` form so the buttons perform the correct actions. This messed up all the styling so I've extracted the styles to a parent div and updated all the other devise forms in order to preserve these changes.

### Side Note

I have such a massive case of dejavu on this. It really feels as though I've fixed it before.

### Further side note

Jira ticket is IJ-151 not IJ-150.

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -22,7 +22,7 @@
     }
   }
 
-  form {
+  .form-container {
     background-color: $black;
     border-radius: 5px;
     padding: 1rem 2rem;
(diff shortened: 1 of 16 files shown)


IeuanSkinner (changes requested):
Seeds conflict needs resolving otherwise it LGTM 👍

IeuanSkinner (approved):

PR #58, opened 2021-04-19
Feature/ij 87 user wellbeing history chart

IeuanSkinner:
Added a series of Wellbeing History Charts to the Team Members User member view. These charts are added to the page as carousel slides so Team Members can navigate between them by using the controls or swiping if they're on a touchscreen device.

The use of carousels prevents the problem of overcrowding from overlapping lines on the charts, I looked into grouping the wellbeing metrics according to categories onto charts so that at most 3-4 lines are on a given chart but this would require significant effort to add and given that Include UK haven't requested it we should consider it an additional piece of work for the time being as this implement suffices.

Screenshot 2021-04-19 at 13 56 15

Screenshot 2021-04-19 at 13 55 55


garethdandrews:
Looking great, the carousel is working really well.

But could we have a faster way to change between the metrics? One way would be to have the metric heading (e.g. housing) be a drop down menu. Or maybe include pagination/navigation dots below the chart, that when hovered (or long press for mobile) show a tooltip for the metric.

garethdandrews:
One minor visual thing I notice, is that the green for 'superb' seems to be lighter than the green for 'great'.

garethdandrews:
Also I wonder if we need to show all of the dates on the x-axis as it looks a little cluttered. I really like the tooltips on hover, so how about reducing the number of x labels?

IeuanSkinner:
> One minor visual thing I notice, is that the green for 'superb' seems to be lighter than the green for 'great'.

The colour scale is taken from the wellbeing assessment chart, if you want to change it you can create a ticket and pick that up but it's not the scope of this ticket.

IeuanSkinner:
> But could we have a faster way to change between the metrics? One way would be to have the metric heading (e.g. housing) be a drop down menu. Or maybe include pagination/navigation dots below the chart, that when hovered (or long press for mobile) show a tooltip for the metric.

Feature requests should go into new tickets not peer review.

IeuanSkinner:
> Also I wonder if we need to show all of the dates on the x-axis as it looks a little cluttered. I really like the tooltips on hover, so how about reducing the number of x labels?

Again this is a feature request and we're also using a 3rd party library to render these charts so asking for changes like this aren't simple things, it requires spending hours reading through API docs just to find a) can this actually be changed and b) how to we even do that. 

Time is better spent sticking with the standard configuration as much as possible and focusing on code we control i.e. we write.

phil-6:
I'm getting a lot of empty space underneath the graph which appears to be caused by: 
```css
#wellbeing-history-chart-wrapper {
    min-height: 95vh;
}
```
removing this seems to fix it.
image

phil-6:
> > One minor visual thing I notice, is that the green for 'superb' seems to be lighter than the green for 'great'.
> 
> The colour scale is taken from the wellbeing assessment chart, if you want to change it you can create a ticket and pick that up but it's not the scope of this ticket.

image


I agree @garethdandrews, please can you create a ticket to "Update Wellbeing Color Scale" with some improved suggested values.

phil-6:
> > But could we have a faster way to change between the metrics? One way would be to have the metric heading (e.g. housing) be a drop down menu. Or maybe include pagination/navigation dots below the chart, that when hovered (or long press for mobile) show a tooltip for the metric.
> 
> Feature requests should go into new tickets not peer review.

I agree with both of you! @garethdandrews please can you add a ticket for this.

IeuanSkinner:
> I'm getting a lot of empty space underneath the graph which appears to be caused by ...

@phil-6 yeah I added this in intentionally so that it preloads that card area for large landscape oriented screens but I'm guessing you're viewing that on a portrait oriented monitor?

IeuanSkinner:
@phil-6 You should be able to add in an orientation property to the media query, can you give this a go and see what the results are? Better off you amending that as I don't have a monitor with which to check the change so it'd be guess work.

phil-6:
@IeuanSkinner Nope, landscape monitor.  I see the issue on whatever width I view it above 750px.
image

phil-6 (changes requested):
Overall I'm happy with this design path, I think Gareth is on to something with an easier way to switch between charts. 

Looking at the back end in more detail, `add_to_history` gets called and makes one database query for each wellbeing assessment. I think its worth looking at this as it can get quite expensive. 
Please can we reduce the number of wellbeing assessments fetched to the last 100, or the last year, whatever is easier to implement, and also look to reduce the number of queries we make on page load

phil-6 (approved):
Thanks for the updates Ieuan, much smoother page load now

I'm happy with the way this looks with less random date too.
image

PR #59, opened 2021-04-19
IJ-150 Devise Emails now send, Admins get emails about Crisis Events and new Team Members

phil-6:
Emails now get sent to admin's when a new crisis is created, when a user updates a crisis and when new team members sign up. All the various devise emails should now send as well. 

Emails are functional but un-styled at the moment. IJ-107 is there to make them pretty.

This is set up with our SendGrid account, you'll need the API key to test locally. For production we will need to set Include up with their own account.


spec/mailers/previews/crisis_event_mailer_preview.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def updated_crisis_email
+    @admin = TeamMember.admins.first
+    @user = User.first
+    @crisis_event = @user.crisis_events.last
+    @crisis_type = @crisis_event.crisis_type.category
+    @crisis_notes = @crisis_event.crisis_notes.includes(:team_member)
+    AdminMailer.updated_crisis_email(@user, @admin, @crisis_event, @crisis_type, @crisis_notes)
+  end
```
IeuanSkinner, line 20: So we only send an email to the first admin in the system?

IeuanSkinner: Also is it definitely only admins who respond to crisis events? 🤔

phil-6: This is just the preview. For testing of emails you can navigate to a page which renders the HTML email. It is nested under spec for this reason.

phil-6: It was my understanding that Crisis Event notifications were just limited to Admins

IeuanSkinner (approved):
LGTM

PR #60, opened 2021-04-19
Contacts Pagination Fix

IeuanSkinner:
Fixed the contacts index using the old pagination setup, updated styling on the new/edit form to match that of the IJ-forms.

Screenshot 2021-04-19 at 16 34 50


phil-6 (approved):
Thanks Ieuan

PR #61, opened 2021-04-20
IJ-142 Back Button Fix

IeuanSkinner:
Back button now leverages the history.back function so the user won't be able to get stuck in loops. I've also fixed a bug with searching the wellbeing assessments where the subheading stats aren't explicitly stating the create_at column on the wellbeing assessments table which it needs to if that table has been joined with the user table as is done when searching.


phil-6 (approved):
LGTM

PR #65, opened 2021-04-21
IJ-152 User Goals

IeuanSkinner:
User can now set short-term and long-term goals on the platform and these goals can belong to any number of goal types (currently set to Aspiration, Hope and Meaning) to better facilitate the 'aim' of the goal. 

Users can also mark the goal as achieved by clicking on it which will mark the goal green and trigger a confetti animation.

Screenshot 2021-04-21 at 14 16 28


phil-6:
Really liking this so far Ieuan!

I have a worry that this page will become extremely bloated over time. 
Do we have any functionality yet to delete goals which may have changed,or archive old completed goals?

IeuanSkinner:
> I have a worry that this page will become extremely bloated over time.
> Do we have any functionality yet to delete goals which may have changed,or archive old completed goals?

There's no requirement specified for that anywhere so no there isn't functionality to archive or delete goals. If you want new functionality you need to raise a new ticket, these things can't keep getting established at the point a pull-request exists as by that point hours of work have already been put in and by having the goalposts move _after_ those hours are put in they become redundant.

The flow needs to be: establish criterion --> implement --> PR, where PR is about engineering concerns i.e. is there a bug introduced? Is there non-performant/sub-optimal code? Are there _minor_ improvements that can be handled now?

phil-6:
> > I have a worry that this page will become extremely bloated over time.
> > Do we have any functionality yet to delete goals which may have changed,or archive old completed goals?
> 
> There's no requirement specified for that anywhere so no there isn't functionality to archive or delete goals. If you want new functionality you need to raise a new ticket, these things can't keep getting established at the point a pull-request exists as by that point hours of work have already been put in and by having the goalposts move _after_ those hours are put in they become redundant.
> 
> The flow needs to be: establish criterion --> implement --> PR, where PR is about engineering concerns i.e. is there a bug introduced? Is there non-performant/sub-optimal code? Are there _minor_ improvements that can be handled now?

I understand what you're saying about feature suggestions on PRs, its not best-practice or ideal.

In an ideal world, a lot more time and consideration would be given to fleshing out tickets before they are worked on, but also in an ideal world we'd have a designer and a PM or BA to do that! We are a small team and we all have ownership over what we build, from the small details to the bigger picture. We all need to be working collaboratively as architect, engineer, designer, product and project manager. 

When we start working on a ticket, we should be considering what the feature is and whether the ticket covers everything that's really needed for it to be a complete feature. If anything else comes to mind as we're working on it we can add it in as we go, or create a new ticket for it to be adding further down the line.

There are going to be times when features get suggested in PRs because they've just been missed when the ticket was being written up, or when while exploring the new work something comes to mind. I don't mind whether this gets tagged on to the end of that particular PR if it feels right to do so or gets spun out into another ticket. 

This does not make the hours already put in redundant. It is rare that the work put in will be wasted, I'll always make efforts to ensure that this isn't the case.


So, for this, do you want to write up a new ticket or would your rather add the extra functionality as part of IJ-152?

PR #66, opened 2021-04-23
Feature/ij 136 team member edit delete notes

garethdandrews:
### Overview

Team members can now edit their notes on a user, updating `content` or `visible_to_user`, and see the edit history of a note (if any). 

user_notes

user_notes_history

The most recent version of the selected note, is highlighted in the brand primary colour, and pinned at the top of the card. A note can be edited here if the current team member is the author and this utilizes the modal.

user_notes_edit

Team members can only edit their own notes. Any team member can see the edit history of a note (if any). The icons on the views respond to these requirements.

### Technicalities

Two fields are added to the notes table, `replacing` and `replaced_by`. These are nullable, self referencing foreign keys which keep track of the edit history, like a doubly linked list.

When updating note `a`, a new note `b` is created. We assign `a.replaced_by(b)` and `b.replacing(a)`. 

This feature is particularly useful as it allows us to determine which notes have an edit history, and quickly determine the edit history of a note.


garethdandrews:
The delete functionality has not been implemented but is up for discussion. 

We have the ability to edit notes and view the edit history, so being able to delete a note from the history doesn't seem right. Neither does deleting the note and its history. It might be something to ask Include, but it seems like they would want to keep all notes for record. 

The users side of things hasn't been implemented yet, but would they be able to see the most recent version of a note, or the whole note history? I would assume the former, and the team members then have the option to toggle `visible_to_user`.

app/controllers/team_members/users_controller.rb
```diff
@@ -16,7 +16,7 @@ def show
       user_location
       wellbeing_assessment
       @note = Note.new
-      @user_notes = @user.notes.includes(:team_member).order(created_at: :desc)
+      @user_notes = @user.notes.includes(:team_member).includes(:replaced_by).order(created_at: :desc)
```
IeuanSkinner, line 19: This can be

`@user.notes.includes(:team_member, :replaced_by).order(created_at: :desc)`

app/javascript/packs/notes.js
```diff
@@ -0,0 +1,23 @@
+const edit_btns = document.querySelectorAll('.edit-btn')
+
+edit_btns.forEach(edit_btn => {
+    edit_btn.addEventListener('click', () => {
+        console.log(edit_btn.dataset.userId)
+        console.log(edit_btn.dataset.noteId);
```
IeuanSkinner, line 6: These can be removed.

IeuanSkinner: In face this whole file can be removed can't it?

app/views/team_members/notes/_edit_modal.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class="modal fade" id="edit-note-modal" tabindex="-1" role="dialog" aria-labelledby="note-modal" aria-hidden="true">
+  <div class="modal-dialog modal-lg col-12" role="document">
+    <div class="modal-content">
+      <%= form_with model: @note, url: user_note_path(@user, @note), method: :put do |form| %>
+        <%= form.hidden_field :id %>
```
IeuanSkinner, line 5: I think this is a redundant field now as the note ID is going to be contained in the url path, this was probably needed when editing a note could be done from the users member page but not now that editing is done from the notes member page.

app/views/team_members/users/show.html.erb
```diff
@@ -196,31 +168,26 @@
...
+              <% end %>
             <% end %>
           <% end %>
         </div>
         <div class="card-footer text-center">
-          <button class="btn btn-primary my-2 py-2 px-5" type="button" data-bs-toggle="modal" data-bs-target="#note-modal">
+          <button class="btn btn-success my-2 py-2 px-5" type="button" data-bs-toggle="modal" data-bs-target="#add-note-modal">
             Add Note
           </button>
+          <%# is there a rails button_to/link_to way to replace this? %>
```
IeuanSkinner, line 181: This can be deleted, the UrlHelpers are only really needed if you're making a request to the back-end whereas this button is purely acting on the front-end to show the add note modal. 

Similar functionality exists in application.html.erb for the SOS Crisis Modal.

db/seeds.rb
```diff
@@ -298,13 +298,29 @@
...
-  Note.create!(
+  note = Note.create!(
     team_member_id: rand(1..TeamMember.count),
-    visible_to_user: true,
+    visible_to_user: [true, false].sample,
     user_id: rand(1..User.count),
     content: Faker::TvShows::TheExpanse.quote
   )
+
+  if (notes_counter % 5).zero?
```
IeuanSkinner, line 310: Can you add the comment: "Every fifth note is replaced by a new note" just so others understand what this is doing 👍

AlexW555 (commented):
I think this screen could use some text just to indicate which of these is the current version. (The yellow works well, just some text like "Current note" then below "Previous versions" 
Screenshot 2021-04-26 at 2 43 47 pm

garethdandrews:
@AlexW555 What do you think about this?

Screenshot from 2021-04-27 13-53-57

app/controllers/team_members/notes_controller.rb
```diff
@@ -1,24 +1,86 @@
...
 
     private
 
+    def nothing_to_update_redirect
+      redirect_to user_path(@user), notice: 'Nothing to update!'
+    end
+
+    def error_redirect
+      redirect_to user_path(@user), flash: { error: 'Something went wrong. Please try again.' }
+    end
```
phil-6, line 55: I really like these. Feels clean and easy to understand. :+1:

app/controllers/team_members/notes_controller.rb
```diff
@@ -1,24 +1,86 @@
...
+    before_action :note_params, only: %i[create update]
 
     # POST /notes
     def create
-      @note = current_team_member.notes.create!(content: note_params[:content],
-                                                visible_to_user: note_params[:visible_to_user],
-                                                user: @user)
-      redirect_to user_path(@user), notice: @note ? 'Note added!' : "The note couldn't be added. Please try again."
+      if (@note = create_note)
+        redirect_to user_path(@user), notice: { flash: 'Successfully added note!' }
```
phil-6, line 12: Could this be `success` instead of `notice` ?

app/controllers/team_members/notes_controller.rb
```diff
@@ -1,24 +1,86 @@
...
+        @user_notes = []
+        @note.chain(@user_notes)
+      end
+    end
+
+    # PUT /notes/:id/update
+    def update
+      nothing_to_update_redirect and return unless @note.changes?(note_params)
+
+      ActiveRecord::Base.transaction do
```
phil-6, line 32: As the Note model extends `ActiveRecord::Base`, could this be more contextual as `Note.transaction`?

app/models/note.rb
```diff
@@ -2,6 +2,25 @@
 class Note < ApplicationRecord
   belongs_to :team_member
   belongs_to :user
+  belongs_to :replaced_by, class_name: 'Note', optional: true, foreign_key: 'replaced_by_id'
+  belongs_to :replacing, class_name: 'Note', optional: true, foreign_key: 'replacing_id'
 
   validates_presence_of :team_member_id, :user_id, :content
+
+  def chain(arr)
```
phil-6, line 10: "arr" is what pirates say, can we use "array" instead please

app/views/team_members/notes/_edit_modal.html.erb
```diff
@@ -0,0 +1,27 @@
...
+            Allow user to see this note?
+            <% end %>
+            <%= form.check_box :visible_to_user, id: 'edit_note_modal_checkbox', class: 'form-check-input' %>
+          </div>
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-primary" data-bs-dismiss="modal">Cancel</button>
+          <%= form.submit 'Update note', class: 'btn btn-success' %>
+        </div>
+      <% end %>
```
phil-6, line 24: I feel as though there is quite a lot of shared code between here and the create modal. Can we do a similar thing to Contacts where the form element is shared between the create and update modals?

app/controllers/team_members/notes_controller.rb
garethdandrews: Good spot.

app/views/team_members/notes/_edit_modal.html.erb
garethdandrews: Yes of course. It was originally split into two separate modals as we were originally allowing notes to be edited from the user member page. So I was using two different forms/modals.

app/controllers/team_members/notes_controller.rb
garethdandrews: Yes you're right. I've tested and it all works!

phil-6 (commented):
I really like this implementation. Great Work Gareth. All my comments are small suggestions to tidy it up a bit. 

From what I can see we still have some files in `views/notes/` are these still in use or are the correct files in `views/team_members/notes/`? I get the feeling the might be orphaned and can be deleted.

app/views/team_members/notes/_edit_modal.html.erb
garethdandrews: We have refactored the two views into a singular modal page, with some challenges with routes along the way. Some cross talk between ID's when using a shared form.

IeuanSkinner:
@garethdandrews can `_add_modal.html.erb` and `_edit_modal.html.erb be` removed now as they're both replaced by the `_modal.html.erb` partial?

garethdandrews:
> @garethdandrews can `_add_modal.html.erb` and `_edit_modal.html.erb` be removed now as they're both replaced by the `_modal.html.erb` partial?

Yeah just noticed this myself 🤦

IeuanSkinner (approved):
LGTM

PR #68, opened 2021-04-26
Ij 122 appointments

AlexW555:
This PR allows matches ticket IJ-122 - Appointments. This is mainly just for feedback on what's been done. 🙂

Data (Model / Migration)
- User - belongs_to
- Team_Member - belongs_to (optional)
- Start - Datetime
- End - Datetime
- Duration - (Calculated Field)
- Where - String
- Who_With - String
- What - String
- Attended - Boolean _(Feature to be implemented)_

User features:
Create new
Show Upcoming, Edit, and Delete
Show Past appointments (Pagninated & Searchable)

Team Member Features:
View all user's appointments
Create new appointments for a user

This PR is so I can check the progress of the ticket and get some feedback! So any ideas or notes would be very appreciated.

Questions needing discussion:
- What do you think of the cards?
- What do you think will be the best way to allow users to mark they attended an appointment? 
- Should team_members be able to edit and delete a user's appointments?
- Should we notify the users if their appointments have been edited?
- Should a user be able to edit an appointment a team_member has created (Currently this is set to no)
- Any design changes?


garethdandrews:
I think the cards are looking good. I would maybe slightly change the ordering, so the most important info is first. (as below)
appointments_pr_card
Also on the 'view all appointments' view, ensure the cards are ordered by date, they seemed to be random.

garethdandrews:
Design changes: I'm wondering if the past appointments don't need to be on the users member page? Maybe on the users member page, show the upcoming appointments, then on the 'view all page' have the two cards for upcoming and past.

garethdandrews:
> Design changes: I'm wondering if the past appointments don't need to be on the users member page? Maybe on the users member page, show the upcoming appointments, then on the 'view all page' have the two cards for upcoming and past.

check out bootstrap card navigation: https://getbootstrap.com/docs/4.0/components/card/#navigation
like in this example: https://codepen.io/cristinaconacel/pen/ePVMME

garethdandrews:
> What do you think will be the best way to allow users to mark they attended an appointment?

Should users mark if they attended, or should team members?

garethdandrews:
Overall though I think this is looking great! The create form is really neat.
Also I think this is a really nice touch.
Screenshot from 2021-04-27 15-32-08

db/seeds.rb
```diff
@@ -49,11 +51,11 @@
   team_member.save!
 end
 
-unless TeamMember.find_by_email('ieuan.skinner@swansea.ac.uk').present?
+unless TeamMember.find_by_email('notieuan.skinner@swansea.ac.uk').present?
```
IeuanSkinner, line 54: Can this be removed?

phil-6:
> > What do you think will be the best way to allow users to mark they attended an appointment?
> 
> Should users mark if they attended, or should team members?

Users. Its important that users have ownership of things like this.

IeuanSkinner:
>     * What do you think of the cards?

I'm not a fan of the varying colours between header and body and there's also some whitespace character being produced between the edit and delete buttons which looks odd in Firefox.

Screenshot 2021-04-28 at 15 04 58

>     * What do you think will be the best way to allow users to mark they attended an appointment?

Add an icon in the top right of the card that appears when an appointment is a past appointment (if it's upcoming they can't possibly have attended) that allows them to mark it attended. I'd suggest this icon: https://fontawesome.com/icons/calendar-check?style=solid 

>     * Should team_members be able to edit and delete a user's appointments?

No, whoever creates the appointment should be able to edit/delete it.

>     * Should we notify the users if their appointments have been edited?

If the appointment was created by them for a team member (i.e. someone at Include UK schedules a meeting with them) and then that team member edits the appointment I think we should send an email to the user but we shouldn't email the user if they're editing appointments they created.

>     * Should a user be able to edit an appointment a team_member has created (Currently this is set to no)

No, see above.

>     * Any design changes?

I've created some on a new branch I'll raise a PR to demonstrate shortly.

IeuanSkinner:
There's a lot the same markup being reused in different views here, can you extract the appointment card into its own partial that can be reused in the different views?

phil-6:
> What do you think of the cards?

This is being addressed in the other PR. This other PR will need to be merged into this branch before this gets merged into main.

> What do you think will be the best way to allow users to mark they attended an appointment?

I think a card footer on the past appointments page. Maybe something like this:
image
Ieuan's suggestion is also good.

> Should team_members be able to edit and delete a user's appointments?
No, it is important that the users have ownership of this

> Should we notify the users if their appointments have been edited?
N/A

> Should a user be able to edit an appointment a team_member has created (Currently this is set to no)
N/A

> Any design changes?
As per other PR


Lots of other good comments above Alex. I think there's a few more steps to go through once you're happy that everything has been covered. Great work so far!

AlexW555:
Thanks all for the feedback! I really appreciate it, I will implement those functional changes and then turn my attention to viewing those design changes (I think Phil is working on them now).

phil-6:
I think we need some additional validation on the create form, it throws an ugly error when the end date is before the start date.
Peek 2021-04-30 10-35

phil-6:
Please can we try and remove the hours from display if the duration is less than 1 hour.
image

phil-6:
When marking an appointment as "attended" page size is not preserved
Peek 2021-04-30 10-52
.

app/controllers/users/appointments_controller.rb
```diff
@@ -0,0 +1,106 @@
...
+      params.require(:appointment).permit(:where, :who_with, :what, :start, :end)
+    end
+
+    def fail(type)
+      redirect_to appointments_path, flash: { error: "#{@appointment.id} #{type} status could not be changed" }
+    end
+
+    def success(status, type)
+      redirect_to appointments_path,
+                  flash: { success: "Appointment is #{status ? 'now' : 'no longer'} #{type}" }
```
phil-6, line 103: I feel as though this is a bit cold at the moment. I feel it should be more celebratory:
"Well done on making that appointment, its now been marked as attended"
"Appointment has been marked as not attended"
or just:

` flash: { success: "#{status ? 'Congrats!' : ' '} Appointment has been marked as #{status ? 'now' : 'no longer'} #{type}" }`

IeuanSkinner:
> I think we need some additional validation on the create form, it throws an ugly error when the end date is before the start date.

The invalid authenticity token isn't to do with start/end date validation but we have now fixed this. Will work on validating the start/end date fields now.

IeuanSkinner (approved):
So far as I'm aware everything is now working with this functionality so I'm good with it 👍

phil-6 (approved):

PR #69, opened 2021-04-26
IJ-160 Added a hub contact hours modal to UserNavbar

AlexW555:
Added a contact hours modal view to the navbar. So users can easily see contact hours of include uk and get access to their contact us page. 

I initially thought this could be in the footer, but let me know any feedback. 🙂
Screenshot 2021-04-26 at 3 43 28 pm

### DIFF
--- a/app/assets/stylesheets/modals.scss
+++ b/app/assets/stylesheets/modals.scss
@@ -6,10 +6,6 @@
     border-style: solid;
     border-image: linear-gradient(to bottom, $danger, $warning) 1 100%;
     text-align: center;
-
-    .modal-title {
-
-    }
   }
 
   .modal-body {
@@ -37,5 +33,19 @@
         float: right;
       }
     }
+  }
+}
+
+#contactModal{
+  a{
+    color: blue;
+  }
+
+  .modal-header{
+    border-image: linear-gradient(to bottom, $brand-primary, $grey) 1 100%;
+  }
+
+  .modal-body {
+    border-image: linear-gradient(to top, $brand-primary, $grey) 1 100%;
   }
 }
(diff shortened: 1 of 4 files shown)


phil-6:
For me, I think this makes more sense to be in the footer, next to the social icons.

IeuanSkinner:
As phil says it's more common for organisations to capture this information in the footer and to be honest given that Include UK already have this on their main site I'm not convinced we need it in this app.

If we do then we need to consider that the team members will probably want to be able to update the information to account for times when they may not be available.

Screenshot 2021-04-28 at 10 02 27

AlexW555:
Screenshot 2021-04-28 at 10 42 13 am
@phil-6 I agree, I did initially think it should be down the bottom but the ticket specified up the top.

I also agree with @IeuanSkinner's point as well. Is it strictly necessary? They are part of include so I would have thought already have their details. Would require additional functionality to update the details if implemented (Ideally now with these new proposed practices, this discussion should have been discussed in the ticket beforehand.)

phil-6:
@AlexW555 @IeuanSkinner 
I don't disagree. But this is what include fed back to us:
image

phil-6:
I think keep what you've done so far but just move the icon to the footer. Maybe add something which says "These are the normal Include Hub opening times, for more up to date information please check on the main website"

AlexW555:
Changes made

app/assets/stylesheets/modals.scss
```diff
@@ -39,3 +35,17 @@
     }
   }
 }
+
+#contactModal{
+  a{
+    color: blue;
```
phil-6, line 41: Please can we use one of the variables here instead of the HTML color?

phil-6 (approved):
One small change but all good.

IeuanSkinner (approved):

PR #70, opened 2021-04-27
IJ-165 Crisis Notes Edit

garethdandrews:
Implemented edit functionality for crisis notes. The same functionality that was introduced for user notes.

Team members can only edit the notes that they created.
If a note has an edit history, this can be viewed by any other team member.

crisis_notes

crisis_notes_history

### Further issues

Pagination isn't currently implemented for crisis notes, is this something to add?

One other thing I note: if you edit and update a note, the update method redirects us back to the same page. But because we are using a modal this causes some issues with the back button, and it takes two clicks to return to the crisis event page  (or more, depending on how many edits were made).
This issue also occurs in the user notes feature.

I thought this line would resolve the issue, but it doesn't seem to do anything.

https://github.com/Legal-Innovation-Lab-Wales/include-journey/blob/aef957846a63eae52e9c5057972b4bbc55236d5a/app/views/team_members/crisis_notes/show.html.erb#L2


config/routes.rb
```diff
@@ -61,7 +61,7 @@
       resources :crisis_events, only: %i[index show] do
         get 'active', action: 'active', on: :collection
         put 'close', action: 'close', on: :member, as: :close
-        resources :notes, only: %i[create show update], :controller => 'crisis_notes', on: :member
+        resources :notes, only: %i[create show update], :controller => 'crisis_notes'
```
IeuanSkinner, line 64: Arrows are old Ruby notation, this can be:

`resources :notes, only: %i[create show update], controller: :crisis_notes`

config/routes.rb
```diff
@@ -61,7 +61,7 @@
...
         # post 'note', action: 'add_note', on: :member, as: :notes
         # get 'show', action: 'show_note', on :member
```
IeuanSkinner, line 66: These can be deleted

config/routes.rb
```diff
@@ -61,7 +61,9 @@
       resources :crisis_events, only: %i[index show] do
         get 'active', action: 'active', on: :collection
         put 'close', action: 'close', on: :member, as: :close
-        post 'note', action: 'add_note', on: :member, as: :notes
+        resources :notes, only: %i[create show update], :controller => 'crisis_notes'
+        # post 'note', action: 'add_note', on: :member, as: :notes
+        # get 'show', action: 'show_note', on :member
```
phil-6, line 66: Commented lines should be removed if they're no longer needed

app/views/team_members/crisis_notes/show.html.erb
```diff
@@ -0,0 +1,27 @@
...
+            <h6 class="text-muted text-center">previous</h6>
+          <% end %>
+          <% @crisis_notes[1..-1].each do |note| %>
+            <%= render 'note_card', note: note, current_note: false, crisis_event_page: false, edit: false %>
+          <% end %>
+        </div>
+      </div>
+    </div>
+  </div>
+  </div>
```
IeuanSkinner, line 24: Redundant closing tag

phil-6:
I'm not sure I like the look of the full yellow card. Would having a yellow border provide sufficient emphasis?

db/seeds.rb
```diff
@@ -314,6 +315,23 @@
     crisis_type_id: rand(1..CrisisType.count)
   )
 
+  crisis_notes_count.times do |i|
+    crisis_note = crisis_event.crisis_notes.create!(
+      team_member_id: rand(1..TeamMember.count),
+      content: Faker::Movies::HarryPotter.quote
+    )
+
+    if (i % 2).zero?
```
IeuanSkinner, line 324: This can actually be `if i.even?` and then RuboCop highlights it can be a guard clause

```
next unless i.even?

new_crisis_note = crisis_event.crisis_notes.create!(
  team_member_id: crisis_note.team_member_id,
  content: Faker::Movies::HarryPotter.quote,
  replacing: crisis_note
)
 
crisis_note.update!(replaced_by: new_crisis_note)
```

db/seeds.rb
```diff
@@ -327,5 +345,8 @@
 puts("Users Created: #{user_counter}")
 puts("Users in Database: #{User.count}")
 
+puts("Contact per User: #{contacts_for_each_user}")
```
IeuanSkinner, line 348: Is outputting this worhwhile? It's a fixed variable at the top of the file right 🤔

garethdandrews: Probably not. We had existing put statements for some other variables so I added them for consistency.

garethdandrews:
> I'm not sure I like the look of the full yellow card. Would having a yellow border provide sufficient emphasis?

@phil-6 

Screenshot 2021-04-28 at 14 48 44

app/controllers/team_members/crisis_notes_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+        @crisis_notes = []
+        @crisis_note.chain(@crisis_notes)
+      end
+    end
+
+    # PUT /crisis_events/:crisis_event_id/notes/:id
+    def update
+      nothing_to_update_redirect and return unless @crisis_note.changes?(crisis_note_params)
+      
+      ActiveRecord::Base.transaction do
```
phil-6, line 32: Is it worth updating this in the same way as the User Notes?

garethdandrews: Absolutely!

phil-6 (approved):
Looks great to me!

PR #71, opened 2021-04-28
Feature/ij 122 appointment style updates

IeuanSkinner:
Here's an example of some style updates I think could improve the appointments from the user view.

Before:
Screenshot 2021-04-28 at 15 04 58

After:
Screenshot 2021-04-28 at 15 04 24

Before:
Screenshot 2021-04-28 at 15 05 39

After:
Screenshot 2021-04-28 at 15 08 29

### DIFF
--- a/app/views/users/appointments/upcoming.html.erb
+++ b/app/views/users/appointments/upcoming.html.erb
@@ -8,26 +8,46 @@
         <% @appointments.each do |appointments| %>
           <div class="col-12 col-md-4">
             <div class="card mt-3 text-al">
-              <div class="card-header bg-primary"><%= appointments.what %> - In
-                <%= distance_of_time_in_words(Time.now, appointments.start) %>
+              <div class="card-header">
+                <span>
+                  <%= appointments.what %> - In <%= distance_of_time_in_words(Time.now, appointments.start) %>
+                </span>
                 <span>
                   <% if appointments.team_member.present? %>
                     <p>Created by <%= appointments.team_member.full_name %></p>
                   <% else %>
-                  <%= link_to edit_appointment_path(appointments.id) do %>
-                    <i class="fas fa-edit" data-toggle="tooltip" data-placement="bottom" title="Edit"></i>
+                    <span>
+                      <%= link_to edit_appointment_path(appointments.id) do %>
+                        <i class="fas fa-edit" data-toggle="tooltip" data-placement="bottom" title="Edit"></i>
+                      <% end %>
+                    </span>
+                    <span>
+                      <%= link_to appointment_path(appointments.id), method: :delete, data: { confirm: 'Are you sure?' } do %>
+                        <i class="fas fa-trash-alt" data-toggle="tooltip" data-placement="bottom" title="Delete"></i>
+                      <% end %>
+                    </span>
                   <% end %>
-                    <%= link_to appointment_path(appointments.id), method: :delete, data: { confirm: 'Are you sure?' } do %>
-                    <i class="fas fa-trash-alt" data-toggle="tooltip" data-placement="bottom" title="Delete"></i>
-                  <% end %>
-                    <% end %>
-                  </span>
+                </span>
               </div>
               <div class="card-body">
-                <p class="card-text"><i class="fas fa-user"></i> <%= appointments.who_with %> </p>
-                <p><i class="fas fa-location-arrow"></i> <%= appointments.where %></p>
-                <p><i class="fas fa-clock"></i> <%= appointments.time_on_date %></p>
-                <p><i class="fas fa-hourglass-start"></i> <%= appointments.duration %></p>
+                <div class="container p-0">
+                  <div class="row">
+                    <div class="col"><i class="fas fa-user"></i></div>
+                    <div class="col"><%= appointments.who_with %></div>
+                  </div>
+                  <div class="row">
+                    <div class="col"><i class="fas fa-location-arrow"></i></div>
+                    <div class="col"><%= appointments.where %></div>
+                  </div>
+                  <div class="row">
+                    <div class="col"><i class="fas fa-clock"></i></div>
+                    <div class="col"><%= appointments.time_on_date %></div>
+                  </div>
+                  <div class="row">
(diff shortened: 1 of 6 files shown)


garethdandrews:
Really like the blue container card for create new appointment. Is this design used anywhere else in the application?

garethdandrews:
I think the dark blue works well for 'create', because the white compliments it. However I don't think it works with the black text on the appointments card

phil-6:
### Initial quick thoughts.
I think the contrast with the Black on Blue is too low. 
I prefer the yellow and grey cards
Globally I still don't really like the grey
I like the blue background to the form, but I think we should maintain consistency throughout the app on forms. Maybe apply that everywhere.

IeuanSkinner:
We currently use the centered form with coloured background on all devise views and the contacts add/edit view (so far as I'm aware).

The grey and yellow don't look good so we'll need to find some compromise, can you also have a play around with the background colour then to find something better?

The CSS rule is in `appointments.scss > .card { background: ... } `

phil-6:
A few ideas for the cards, I think we need to keep the styling of these small cards the same. So changes here are reflected on contact cards as well.
### 1 
image

### 2
image

### 3
```scss
      .card-header {
        background: linear-gradient(to right, $brand-secondary-alt-lightest, $brand-complementary-lightest, $brand-primary);
        font-weight: 700;
        font-family: $font-link;
        color: $brand-primary;
```
image

### 4
``` scss
.card-header {
        background: linear-gradient(to right, $brand-secondary-alt-dark, $brand-secondary-alt,  $brand-complementary-lightest);
        font-weight: 700;
        font-family: $font-link;
        color: $brand-primary;
```
image

### 5
```scss
      .card-header {
        background: linear-gradient(to right, $brand-complementary, $brand-secondary-alt-darkest, $brand-secondary-alt);
        font-weight: 700;
        font-family: $font-link;
        color: $brand-primary;

        span:last-child{
          float: right;

          span {
            &:last-child {
              margin-left: 0.5rem;
            }
            i.fa-trash-alt {
              color: $danger;
            }
            i.fa-edit {
              color: $info;
            }
            a:hover i {
              color: $primary;
            }
          }
        }
      }
```
image


## Summary
So, a few different options there. We can definitely pick and choose which aspects of which ones we want. What do you think?
@IeuanSkinner @garethdandrews @AlexW555 @bmhtech07

IeuanSkinner:
The gradient colouring looks very Microsoft Word clip-art to me, not unappealing but not particularly professional either, having said that of the above I think 3 is the prettiest.  

I think the wider problem might be the overreliance of bootstrap cards in the platforms UI (which is largely the result of chop & swap and not uniqute thinking for each view).

How is the header dealing with longer entries for the 'what' field? It's not unlikely that people might use a sentence or 2 to explain _what_ the appointment is about.

phil-6:
> Microsoft Word clip-art to me

image

>  not particularly professional either
While I fundamentally disagree with this, does this app have to be "professional"

> I think the wider problem might be the overreliance of bootstrap cards in the platforms UI (which is largely the result of chop & swap and not unique thinking for each view).

Yeah, I agree, the discussion around the Team Member user view is valuable from moving away from this over-reliance. Maybe it makes more sense to have these as a single column. I'll have a play.

phil-6:
Still cards, but one per row.

image

IeuanSkinner:
@phil-6 Honestly I have no idea, I'm completely out of depth deciding what is good UI/UX especially in the absence of any reference material.

garethdandrews:
> Still cards, but one per row.
> 
> image

@phil-6 I think you're on to something here. I think the solid block of cards is a lot to take in, we should add something to break up the list. Maybe grouping the appointments by month? Take a look at how outlook does it below.

Screenshot 2021-04-29 at 14 07 28

garethdandrews:
Whilst I agree that we don't necessarily need it to look at feel highly professional, I don't particularly like gradients. On top of that, they may cause some accessibility issues with the contrasts.

I do however like the '$brand-secondary-alt-lightest' with the yellow text in the header in number 3.

phil-6:
Gradients have a time and a place. I'm not saying these gradients are the right gradients, but I **MUCH** prefer this to the plain flat grey. 

I've also grouped the appointments by day now.

image

garethdandrews:
I'm really happy with the way this is looking.
> Gradients have a time and a place.
This simple gradient works really well, the busy multi coloured gradients were too much.

app/views/users/appointments/upcoming.html.erb
```diff
@@ -5,30 +5,55 @@
     <h3 class="mb-0">Upcoming appointments</h3>
     <div class="container">
       <div class="row">
-        <% @appointments.each do |appointments| %>
-          <div class="col-12 col-md-4">
+        <% @appointments.group_by { |t| t.start.day }.each do |appointments| %>
```
IeuanSkinner, line 8: This is awesome! 😮

IeuanSkinner:
@phil-6 I'm happy for this to be merged 👍

app/views/users/appointments/upcoming.html.erb
phil-6: It was slightly wrong, updated it so it should be more reliable now - grouped by date now, not just number-of-day-of-the-month!

phil-6:
Updated the create form. 
Subheading text is now white and background is now the same gradient as the heading on the appointment groups

image

@garethdandrews @IeuanSkinner  @AlexW555  Too much gradient or is this the Goldilocks amount?

IeuanSkinner:
@phil-6 looks fine to me 👍

phil-6:
So, I got a little carried away, and made some changes to other areas of the app in the same theme as above.
I really didn't like that grey!

Log in cards:
image

User Index cards:
image

Some of the changes I've made affect the Team Member User View but I'm not making changes there due to the ongoing conversation with IJ-166. 

Wellbeing assessment cards:
image

This is definitely still a work in progress. But does everyone like this direction?

garethdandrews:
LGTM!

phil-6 (approved):

PR #75, opened 2021-04-29
Feature - IJ-102 Team Members: Crisis Events Updates

bmhtech07:
Make changes per IJ-102 to the Crisis Events displays (both open and closed events).

Refactored some of the crisis_events_controller.


app/controllers/team_members/crisis_events_controller.rb
```diff
@@ -1,12 +1,13 @@
...
   class CrisisEventsController < TeamMembersApplicationController
-    include Pagination
     before_action :crisis_event, except: %i[index active]
+    before_action :crisis_events, :set_variables, only: %i[index active]
+    include Pagination
 
     # GET /crisis_events/active
     def active
-      @crisis_events = CrisisEvent.active.includes(:user, :crisis_type).order(updated_at: :desc)
+      crisis_events
```
garethdandrews, line 10: This line can be removed as you have it in a before action

garethdandrews, line 10: This line can be removed, as you already make the method call in a before action

app/controllers/team_members/crisis_events_controller.rb
```diff
@@ -44,16 +45,25 @@ def search
 
     def crisis_event
       @crisis_event = CrisisEvent.includes(:user, :crisis_type).find(params[:id])
-    rescue ActiveRecord::RecordNotFound
-      redirect_back(fallback_location: active_crisis_events_path, flash: { error: 'Crisis event not found' })
```
phil-6, line 48: Why are we removing this error handling?

app/controllers/team_members/crisis_events_controller.rb
```diff
@@ -44,16 +45,25 @@ def search
...
+    end
+
+    def crisis_events
+      @crisis_events = CrisisEvent.active.includes(:user, :crisis_type, :crisis_notes).order(updated_at: :desc)
+    end
+
+    def set_variables
```
phil-6, line 54: I like this as one method, would `set_statistics` be more descriptive?

app/views/team_members/crisis_events/active.html.erb
```diff
@@ -1,16 +1,29 @@
...
         <div class="table-container">
-          <table class="table table-dark table-striped link-table">
+          <table class="table table-dark table-striped link-table table-hover">
             <thead>
               <th scope="col">User</th>
               <th scope="col">Crisis Type</th>
-              <th scope="col">Additional Info</th>
-              <th scope="col">Last Update</th>
+              <th scope="col">Latest Update</th>
+              <th scope="col">Updated</th>
```
garethdandrews, line 26: I much prefer this!

app/controllers/team_members/crisis_events_controller.rb
bmhtech07: I think in error. I don't remember actually stripping that out. I'll have a look

bmhtech07:
Changes made per first review - review required

phil-6 (approved):

app/controllers/team_members/crisis_events_controller.rb
```diff
@@ -48,12 +47,23 @@ def crisis_event
       redirect_back(fallback_location: active_crisis_events_path, flash: { error: 'Crisis event not found' })
     end
 
+    def crisis_events
+      @crisis_events = CrisisEvent.active.includes(:user, :crisis_type, :crisis_notes).order(updated_at: :desc)
+    end
+
+    def set_overview_stats
+      @crisis_events_users = @crisis_events.distinct.count(:user_id)
+      @closed_events = @crisis_events.closed
```
IeuanSkinner, line 56: This doesn't work, `@crisis_events` are all crisis events that are active (i.e. not closed) so looking for closed crisis events on them will always yield 0.

Screenshot 2021-04-30 at 12 04 35

app/controllers/team_members/crisis_events_controller.rb
```diff
@@ -48,12 +47,23 @@ def crisis_event
...
+    end
+
     def crisis_search
       'lower(crisis_types.category) similar to lower(:query) or lower(additional_info) similar to lower(:query)'
     end
 
-    def subheading_stats
-      # TODO: Add stats for closed crisis events index
+    def crisis_notes_params
+      params.require(:crisis_note).permit(:content)
```
IeuanSkinner, line 66: This function is from the `CrisisNotesController` it shouldn't be in the `CrisisEventsController`.

app/views/team_members/crisis_events/active.html.erb
```diff
@@ -1,16 +1,29 @@
...
       <% if @crisis_events.count > 0 %>
         <div class="table-container">
-          <table class="table table-dark table-striped link-table">
+          <table class="table table-dark table-striped link-table table-hover">
             <thead>
               <th scope="col">User</th>
               <th scope="col">Crisis Type</th>
-              <th scope="col">Additional Info</th>
-              <th scope="col">Last Update</th>
+              <th scope="col">Latest Update</th>
```
IeuanSkinner, line 25: "Latest Note" would be more accurate

app/views/team_members/crisis_events/show.html.erb
```diff
@@ -1,5 +1,5 @@
 <main class="container team-members crisis-event">
-  <%= render 'shared/back_btn' %>
+  <%= render 'shared/back_btn', path: active_crisis_events_path, text: 'Active Crisis Events' %>
```
IeuanSkinner, line 2: `back_btn` no longer has a `path` and `text` field, it now functions as the browser back button does so these can be deleted.

app/views/team_members/crisis_events/show.html.erb
```diff
@@ -35,9 +35,14 @@
           <div class="card-footer">
             <div class="container p-0">
               <div class="row">
-                <div class="col-6">
-                  <a class="btn btn-primary p-2" data-bs-toggle="modal" data-bs-target="#note-modal">New Note <i class="fas fa-plus-circle"></i></a>
-                </div>
+                <% if @crisis_event.closed && @crisis_event.closed_at < 7.days.ago %>
+                  <span>Notes can't be added to Crisis Events closed over 7 days ago.</span>
```
IeuanSkinner, line 39: This isn't a constraint that the back-end enforces and if you have a note you created on a crisis event that was closed over 7 days ago you can still edit that note which is, in principal, adding a new note to that crisis event.

app/controllers/team_members/crisis_events_controller.rb
bmhtech07: Thankyou, good spot!

app/views/team_members/crisis_events/active.html.erb
bmhtech07: Not quite, as the 'update' could well just be the `crisis_event.additional_info`

IeuanSkinner: Crisis notes are created by Team Members whereas the Additional Info for a crisis event is added by the user, if we're mix and matching between these 2 pieces of information how is a team member to know looking at this view what the origin of this text is? 

Another problem is we're saying "Latest Update" but that's not strictly speaking true as the branching logic in the corresponding table cell doesn't compare the timestamp of the latest note and the timestamp for the latest update to the additional info field instead it will display the last note if any notes are present otherwise it falls back to the additional info meaning the additional info could have been updated since the last note was written but it won't show up here.

The simplest path forward I can see for both of these is to have a column for the additional info and a column for the "Latest Note" which the ticket description seems to ask for however we're running the risk of displaying too much unnecessary information though, if someone wants to look at the latest note I would've thought the sensible action is to click into and view the crisis event record.

app/controllers/team_members/crisis_events_controller.rb
```diff
@@ -48,12 +48,20 @@ def crisis_event
...
+      @active_events = @crisis_events.active
     end
 
-    def subheading_stats
-      # TODO: Add stats for closed crisis events index
+    def set_overview_stats
+      @crisis_events_users = @active_events.distinct.count(:user_id)
+      @closed_events = @crisis_events.closed
+      @closed_events_30_days = @closed_events.where("closed_at >= ?", 30.days.ago)
+      @closed_events_7_days = @closed_events.where("closed_at >= ?", 7.days.ago)
```
IeuanSkinner, line 60: If not using string interpolation single quotes should be used.

app/controllers/team_members/crisis_events_controller.rb
IeuanSkinner: The next problem here is we're now loading in the `closed` crisis events on the active index and similarly we're loading the `active` crisis events on the closed index which defeats a lot of the reasons for having these be seperate views in the first place. 

We're also loading in the `closed` crisis events twice on the closed index page, the first time we're loading them in as `@resources` via the pagination concern and now this code block is loading them in again. You can leverage the call to `subheading_stats` that the pagination concern makes to  access these stats from the `@resources` instance variable before the index is loaded i.e. how `wellbeing_assessments_controller` works. This will also mean search queries will be reflected in these stats i.e. if there's 100 closed crisis events and 5 raised by "John" then when I search "John" the current setup will keep saying 100 despite the fact that doing a search is a form of filtering the collection, if instead you access `@resources` then this value of 5 will come through.

Can we load and display information relating to the closed crisis events on the closed crisis events index page?

app/views/team_members/crisis_events/show.html.erb
phil-6: What are you suggesting here?

app/views/team_members/crisis_events/active.html.erb
phil-6: Maybe just inspect the additional info column and have "Latest User Update"?

app/views/team_members/crisis_events/show.html.erb
IeuanSkinner: If we can't add a new note that's over 7 days old we probably shouldn't be able to edit them as editing is just another form of creating. We also need to make sure that the back-end is the main enforcer of this rule not just the front-end i.e. don't just remove the controls but have a check on the back-end's `create`/`update` actions to see if the age of the crisis event is over 7 days and if so do not create the record.

app/views/team_members/crisis_events/active.html.erb
bmhtech07: I've changed this to show the latest of additional info or latest note.

I've added a badge to indicate it's a note. I feel it's succinct enough  to give a high level overview of the latest info, without overloading with info

image

IeuanSkinner: Could we maybe have an "Info" badge too so they all align? Also the white on yellow is a bit tough to read so perhaps black would work better there.

app/views/team_members/crisis_events/show.html.erb
IeuanSkinner: It _can_ be its own ticket but then I'd expect that ticket to be about removing the front-end controls and enforcing the back-end constraint rather than have one part of the work done in this ticket and the other in another.

app/views/team_members/crisis_events/active.html.erb
bmhtech07: image

IeuanSkinner: How does it look with the column center aligned or all the columns left aligned? The mix-n-match between them looks slightly odd.

IeuanSkinner:
As an alternative we can center the Latest Update text whilst having the badge left-aligned, thoughts? @phil-6 

Screenshot 2021-05-06 at 12 02 19

phil-6:
I prefer the aligned and centered version. Could we move the `Type` column to after the `Lastest Update` col and rename it `Update Type`?

garethdandrews:
I prefer version two - either everything needs to be left or center aligned, not a mix. And I agree with what Phil is saying about the columns.

IeuanSkinner:
@phil-6 , @garethdandrews I've now made this change 👍 

Screenshot 2021-05-10 at 15 56 01

PR #78, opened 2021-04-30
Ij 163 wellbeing history chart navigation

garethdandrews:
Implemented another way of navigating between the wellbeing history chart carousel. The chart title has been replaced with a select drop-down list and indicators have been added below the chart to assist with navigation.

Users can navigate using the select menu, carousel controls (arrows) or the indicator buttons. The select menu and indicators react and update accordingly.

Additionally, I have reduced the number of ticks on the x-axis as I don't believe all the date points are necessary. The x-axis should provide a general overview, and hovering a specific point should reveal more specific details.

Before:
wellbeing_history_chart_before

After:
wellbeing_history_chart_navigation


IeuanSkinner:
I've reduced the points size on the chart as I think it looks a bit cleaner, thoughts?

Screenshot 2021-04-30 at 12 41 38

garethdandrews:
@IeuanSkinner yeah I like that, it looks a lot cleaner :+1:

IeuanSkinner (approved):
LGTM

PR #79, opened 2021-04-30
If the user has no contacts, then redirect to 'add contact' form

garethdandrews:
### Problem
Error was thrown if contacts button was clicked on dashboard when user has no contacts.

### Solution
If the user has no contacts, they are redirected to the add contact view. Otherwise, they are redirected as normal.

### DIFF
--- a/app/views/users/dashboard/show.html.erb
+++ b/app/views/users/dashboard/show.html.erb
@@ -12,8 +12,14 @@
                      large_text: 'My Needs', text: 'Needs' %>
         </div>
         <div class="row">
-          <%= render 'shared/dashboard_btn', path: contacts_path, icon: 'fa-address-book',
+          <% if current_user.contacts.count.zero? %>
+            <%= render 'shared/dashboard_btn', path: new_contact_path, icon: 'fa-address-book',
                      large_text: 'My Contacts', text: 'Contacts' %>
+          <% else %>
+            <%= render 'shared/dashboard_btn', path: contacts_path, icon: 'fa-address-book',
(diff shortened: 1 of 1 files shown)


app/views/users/dashboard/show.html.erb
```diff
@@ -12,8 +12,14 @@
...
         </div>
         <div class="row">
-          <%= render 'shared/dashboard_btn', path: contacts_path, icon: 'fa-address-book',
+          <% if current_user.contacts.count.zero? %>
+            <%= render 'shared/dashboard_btn', path: new_contact_path, icon: 'fa-address-book',
                      large_text: 'My Contacts', text: 'Contacts' %>
+          <% else %>
+            <%= render 'shared/dashboard_btn', path: contacts_path, icon: 'fa-address-book',
+                     large_text: 'My Contacts', text: 'Contacts' %>
+          <% end %>
```
IeuanSkinner, line 21: This can be made even shorter with the power of ternary statements

```
<%= render 'shared/dashboard_btn', path: current_user.contacts.count.zero? ? new_contact_path : contacts_path,
                     icon: 'fa-address-book', large_text: 'My Contacts', text: 'Contacts' %>
```

garethdandrews: Thanks for pointing this out, don't know how I didn't spot that!

garethdandrews:
Upon reflection, I don't think you can use the ternary operator in this way. Rails throws a ActionView::SyntaxErrorInTemplate

PR #80, opened 2021-05-04
Feature - IJ-119 Journal Entry View Logs Sort

IeuanSkinner:
Team Member's Journal Entry View Log Index page is now sortable by Created At Ascending/Descending and Viewed At Ascending/Descending.

Screenshot 2021-05-04 at 12 23 18

### DIFF
--- a/app/controllers/team_members/journal_entry_view_logs_controller.rb
+++ b/app/controllers/team_members/journal_entry_view_logs_controller.rb
@@ -8,12 +8,15 @@ module TeamMembers
 
     def resources
       @team_member.journal_entry_view_logs.includes(:user, :journal_entry)
+                  .joins(:journal_entry)
+                  .order(sort)
     end
 
     def search
       @team_member.journal_entry_view_logs.includes(:user, :journal_entry)
-                  .joins(:user)
+                  .joins(:journal_entry)
                   .where(user_search, wildcard_query)
+                  .order(sort)
     end
 
     def team_member
@@ -23,6 +26,21 @@ module TeamMembers
     def subheading_stats
       # TODO: Add stats for team member view logs index
 
+    end
+
+    def sort
+      sort_param = journal_entry_view_logs_params[:sort]
+
+      if %w[created_at_asc created_at_desc viewed_at_asc viewed_at_desc].include?(sort_param)
+        sort_param = sort_param.split('_')
+        { "#{sort_param[0..1].join('_') == 'viewed_at' ? 'created_at' : 'journal_entries.created_at'}": sort_param[2] }
+      else
+        { 'created_at': :desc }
+      end
+    end
+
+    def journal_entry_view_logs_params
+      params.permit(:page, :query, :limit, :sort)
     end
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/team_members/journal_entry_view_logs_controller.rb
```diff
@@ -24,5 +27,20 @@ def subheading_stats
       # TODO: Add stats for team member view logs index
 
     end
+
+    def sort
+      sort_param = journal_entry_view_logs_params[:sort]
+
+      if %w[created_at_asc created_at_desc viewed_at_asc viewed_at_desc].include?(sort_param)
+        sort_param = sort_param.split('_')
+        { "#{sort_param[0..1].join('_') == 'viewed_at' ? 'created_at' : 'journal_entries.created_at'}": sort_param[2] }
```
garethdandrews, line 36: Could use Regex here.
`/_[^_]*$/`

IeuanSkinner: What's the justification too? The sort param needs to conform to the whitelist in the conditional wrapping it so we know the format is always going to be `some_field_direction` where `some_field` and `direction` have clear indexes.

If I were too see these rules in the code I'd assume it meant that we allowed a range of values that only needed to conform to the string structure as defined in the regex i.e. `this_value_is_allowed` but since we're already filtering the options to a whitelist applying regexp that are broader than the scope of that whitelist seems odd.  If we wanted to use regexp why not just use `^(created_at|viewed_at)_(asc|desc)$`?

My reason for not using the above regexp would be that it's much easier to read an inline whitelist of `[created_at_asc created_at_desc viewed_at_asc viewed_at_desc]` than read and interpret that regexp despite them meaning the same thing.

garethdandrews: When I first read this, I assumed you were just trying to determine `some_field` and `direction`, and return them as a block, so it didn't make sense why you would split a string then re-join some part of it. I thought 
`{ "#{sort_param[0..1].join('_') == 'viewed_at' ? 'created_at' : 'journal_entries.created_at'}": sort_param[2] }`
could be replaced with
`{ /.*(?=\_)/.match(sort_param'): /[^_]*$/.match(sort_param) }`
to extract the necessary parts.
However with the conditional statement it may begin to look more complicated than I originally thought.

I agree with your justification for the whitelist, my intention for the regex was to replace the `split` and `join`:

garethdandrews: So I thought you could do something like:
`if %w[created_at_asc created_at_desc viewed_at_asc viewed_at_desc].include?(sort_param)`
`some_field = /.*(?=\_)/.match(sort_param)`
`direction = /[^_]*$/.match(sort_param)`
`{ "some_field == 'viewed_at' ? 'created_at' : 'journal_entries.created_at'}": direction }`

In my head I thought this would be more readable, but looking at it.. maybe not

phil-6 (approved):
Looks good to me. Still impressed with how easily your pagination structure makes this extensible.

PR #81, opened 2021-05-04
Feature - IJ-123 User Journals Dashboard Updates

garethdandrews:
Removed the Journal Entries dashboard, and redirect users to 'add new' page if no journal entries are present. Added subheading stats to journal entries index.

I've got two buttons to add a new journal entries. Let me know which one I should keep/remove.

journal_entries_index


IeuanSkinner:
Not a fan of having 2 buttons to do the same function within the same view (i.e. the big square new entry button and the smaller green new entry button). My preference is to keep the green button and lose the big square one.

garethdandrews:
> Not a fan of having 2 buttons to do the same function within the same view (i.e. the big square new entry button and the smaller green new entry button). My preference is to keep the green button and lose the big square one.

@IeuanSkinner The plan is to remove one of them. I couldn't decide which I preferred.

phil-6:
On Buttons:
I don't think either is quite right at the moment. 
I think my preference would be smaller "primary" button, directly under the title.

app/assets/stylesheets/buttons.scss
```diff
@@ -87,5 +87,9 @@ body#overboot {
     &.two {
       max-width: 400px;
     }
+
+    &.one {
+      max-width: 160px;
+    }
```
phil-6, line 93: Can this go above "two" please - if its still needed.

phil-6 (commented):
LGTM, just the button changes. :+1:

garethdandrews:
Style updates as per comments

AlexW555 (approved):
LGTM: In regards to the button, there isn't a consistent style for where an "add new" button should go or look like. But I like the green button look here

IeuanSkinner:
@garethdandrews when you update the UI in later commits can you be sure to include screenshots 👍

PR #82, opened 2021-05-05
Feature - IJ-118 Team Members: Journal Index Search

IeuanSkinner:
I've upgraded the search controls on the team members journal index so that the viewed/not viewed status of a journal entries can also be searched for as can the feeling (emoji) that the user associated with that journey.

Screenshot 2021-05-05 at 09 34 58


garethdandrews (commented):
LGTM!

AlexW555 (approved):
👍🏻Brilliant!

PR #84, opened 2021-05-05
Feature - IJ-158 Team Member: User Profile View Logs

garethdandrews:
Implemented view logs feature for user profiles. 

A new button has been added to the team member show page to access this new page.

Screenshot 2021-05-05 at 15 37 18

A log is produced every time a team member views a users profile., similarly to how the journal entries view log is generated.

Screenshot 2021-05-05 at 15 15 46

### DIFF
--- a/app/controllers/team_members/users_controller.rb
+++ b/app/controllers/team_members/users_controller.rb
@@ -13,6 +13,7 @@ module TeamMembers
 
     # GET /users/:id
     def show
+      view_log
       user_location
       wellbeing_assessment
       @note = Note.new
@@ -79,6 +80,12 @@ module TeamMembers
     end
 
     private
+
+    def view_log
+      return if current_team_member.user_profile_view_logs.create!(user: @user)
+
+      redirect_back(fallback_location: authenticated_team_member_root_path, alert: 'View log could not be created')
+    end
 
     def user
       @user = User.includes(:notes).find(params[:id])
--- a/spec/factories/user_profile_view_logs.rb
+++ b/spec/factories/user_profile_view_logs.rb
@@ -1,0 +1,5 @@
+FactoryBot.define do
+  factory :user_profile_view_log do
+    
+  end
+end
--- a/app/assets/stylesheets/user_profile_view_logs.scss
+++ b/app/assets/stylesheets/user_profile_view_logs.scss
@@ -1,0 +1,10 @@
+.user-profile-view-logs {
+  .fa-long-arrow-alt-right {
+    font-size: 2.5rem;
+  }
+
+  .date-sort-select {
+    width: 240px !important;
+    float: right;
+  }
+}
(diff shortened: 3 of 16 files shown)


IeuanSkinner:
In terms of storage capacity every view log is a row in a table and each row contains the lookup to the resource id (journal or user in this case) and the team member who viewed that resource and so for repeat viewings of a resource we're getting redundant info i.e. for every time a journal entry is viewed we repeat 2 of the 4 rows for it, same thing here we'd be repeating the user and team member id for every view.

I'm less bothered with this w.r.t. journal entries as duplication is dependant on repeat viewings of the resource and this isn't that likely to happen but the user profile is completely different as it's very likely that a team member will look at a users profile multiple times.

As an alternative we can create a single entry in the view log table and track a count for the views, this way we would know the first time they viewed the profile via `created_at`, the last time they viewed the profile via `updated_at`, the number of times they viewed the profile by `view_count` and we can only ever have a maximum number of rows of n*m where n is the number of team members and m is the number of users.

garethdandrews:
Changes made as per comments.
Updated the view to include view_count:

user_profile_view_logs_view_count

app/controllers/team_members/users_controller.rb
```diff
@@ -80,6 +81,15 @@ def search
 
     private
 
+    def log_view
+      UserProfileViewLog.transaction do
+        view_log = current_team_member.user_profile_view_logs.find_or_create_by!(user: @user)
+        view_log.save!
+      end
+    rescue ActiveRecord::RecordInvalid
```
IeuanSkinner, line 89: Does this need to be wrapped in a transaction? I suspect this is actually doing some redundant operations as it's creating a local variable that it needn't.

```
def log_view
    current_team_member.user_profile_view_logs.find_or_create_by!(user: @user)    
rescue ActiveRecord::RecordInvalid
      redirect_back(fallback_location: authenticated_team_member_root_path, flash: { error: 'View log could not be created' })
    end
```

spec/factories/user_profile_view_logs.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :user_profile_view_log do
+    
+  end
+end
```
IeuanSkinner, line 5: These boiler plate files under `spec` can be deleted.

app/controllers/team_members/users_controller.rb
garethdandrews: Not sure what I was thinking there?

garethdandrews:
Updated the order selector and column headers as per review.

Screenshot from 2021-05-11 11-09-50

garethdandrews:
This view also needs to be updated to use the new searchable layout once it is pushed to main

IeuanSkinner:
Your default is 'created_at_desc' but the first item in your select dropdown is 'created_at_asc' so if you view this index with no sort parameter then the sort being applied and the option shown in the dropdown don't match.

Screenshot 2021-05-11 at 13 04 48

garethdandrews:
I've fixed that mismatch

PR #88, opened 2021-05-11
Ij 145 subheading stats

AlexW555:
Implemented subheading stats for the remaining pages not already covered by other branches. [IJ-145](https://lilw.atlassian.net/browse/IJ-145?atlOrigin=eyJpIjoiODllYWMyOTk3MGM0NDdkNGExMTRjY2YzNWY0YmFlZDgiLCJwIjoiaiJ9)

### **TeamMembers** 
- Wellbeing assessments
Screenshot 2021-05-11 at 1 15 21 pm

 - Appointments
 Screenshot 2021-05-11 at 1 14 36 pm

### **Users**  
- Appointments 

Screenshot 2021-05-11 at 1 15 55 pm


IeuanSkinner:
I'm not an English grad. so could be wrong but I think you need to use 'are' rather than 'is' when the noun is plural so it should be "There are a total of 8 appointments" etc.

app/controllers/team_members/appointments_controller.rb
```diff
@@ -63,5 +63,15 @@ def past_appointment_query
...
       params.require(:appointment).permit(:where, :who_with, :what, :start, :end)
     end
+
+    def subheading_stats
+      @count_in_last_week = @resources.where('appointments.start >= ?', 1.week.ago).size
+      @count_in_last_month = @resources.where('appointments.start >= ?', 1.month.ago).size
+      @count_in_next_weeks = @resources.where('appointments.start <= ?', 2.week.ago).size
+      return unless @user
+
+      @count_by_team_member = @resources.count { |appointment| appointment.team_member_id.present? }
```
IeuanSkinner, line 73: These instance variables aren't being used?

IeuanSkinner: Also the `return unless @user` line doesn't make sense as `@user` is set as a `before_action` so it'll always be present and thus never return.

app/controllers/users/appointments_controller.rb
```diff
@@ -113,5 +113,10 @@ def validate_dates
 
       redirect_back(fallback_location: new_appointment_path, flash: { error: 'End date cannot be before start date' })
     end
+
+    def subheading_stats
+      @count_in_last_week = @resources.where('appointments.start >= ?', 1.week.ago).size
+      @count_in_last_month = @resources.where('appointments.start >= ?', 1.month.ago).size
```
IeuanSkinner, line 119: Same comments apply here as apply in the `TeamMember` version of the controller.

app/views/team_members/appointments/index.html.erb
```diff
@@ -25,7 +31,7 @@
                 </div>
                 <div class="card-footer d-flex"><span class="me-auto">Attended: </span>
                   <div class="btn disabled <%= appointment.attended ? 'btn-success' : 'btn-danger' %>">
-                    <% if appointment.attended ? 'btn-success' : 'btn-danger'%>
+                    <% if appointment.attended %>
```
IeuanSkinner, line 34: Why is this removed?

IeuanSkinner:
@AlexW555 where the 'searchable' layout is used the stats need to be inserted into the view by using a partial that is passed to the template. If you checkout the main branch and take a look at the `team members > crisis index`, `team members > wellbeing assessments index` and `users > journal entries index` these should all provide insight on how this works.

app/views/team_members/appointments/index.html.erb
AlexW555: I thought it was covered by the line above?
`<div class="btn disabled <%= appointment.attended ? 'btn-success' : 'btn-danger' %>">`

garethdandrews: Yeah line 28 is redundant, it's checking the same thing.

app/views/team_members/wellbeing_assessments/index.html.erb
```diff
@@ -83,6 +83,11 @@
           <h3><%= @team_member.full_name %> wellbeing assessments</h3>
         <% else %>
           <h3>All wellbeing assessments</h3>
+          <div class="subheading">
+            <p>There is a total of <%= @count %> wellbeing assessments.</p>
```
garethdandrews, line 87: Is it worth pluralizing this count in the rare case that there is 1 wellbeing assessment?

garethdandrews:
It's all looking really good! A few minor tweaks and merge conflicts to resolve and it's done :)

IeuanSkinner (approved):
LGTM 👍

garethdandrews (approved):
LGTM, good work guys!

PR #91, opened 2021-05-13
Feature - IJ-171 Handle User Account Deletions

IeuanSkinner:
I've renamed the 'pause' functionality for Team Members to 'suspended', we might want to consider partitioning suspended Team Members into a seperate view from the non-suspended Team Members as the reason to suspend a Team Member is typically going to be things like them not being with the organisation anymore so they're in effect archived.

For users I've added a new field to the User model called 'Deletion Date' which is a timestamp that indicates when the User account should be deleted, the deletion is setup to cascade to remove all the linked data with the User including:

```
User > Wellbeing Assessments > Wellbeing Assessment Scores
         > Journal Entries > Journal Entry Permissions > Journal Entry View Logs
         > Goals
         > Appointments
         > User Profile View Logs
         > Notes
         > Crisis Events > Crisis Notes
```

This will ensure there are no orphan records, if you think there's anything I've missed let me know.

**Further details instructions on this functionality are in the README file.**

---

When a user triggers the account deletion via the 'Cancel Account' button on the 'Edit Profile' view the deletion date field will be set and a banner will appear at the top of their screen indicating the countdown to their account being deleted. They can undo this action by clicking on the button indicated which will set the deletion date field back to null.

Desktop:
Screenshot 2021-05-13 at 13 14 48

Mobile:
Screenshot 2021-05-13 at 13 16 40


garethdandrews:
I think this looks great, good job!

garethdandrews (approved):

app/controllers/application_controller.rb
```diff
@@ -9,4 +10,29 @@ def configure_permitted_parameters
     devise_parameter_sanitizer.permit :sign_up, keys: added_attrs
     devise_parameter_sanitizer.permit :account_update, keys: added_attrs
   end
+
+  def deletion_date
+    return unless user_signed_in?
```
phil-6, line 15: Is this redundant due to the conditional on the before_action filter?

phil-6 (approved):
LGTM!

app/controllers/application_controller.rb
IeuanSkinner: It is I've now removed it.

PR #92, opened 2021-05-13
Feature/IJ-173 Restructure Journal Entry View Logs

garethdandrews:
Restructured `JournalEntryViewLogs`, in a similar way to the behaviour of `UserProfileViewLogs`. Before, if the user refreshes the page repeatedly it would create a new view log for each view, which would eventually bloat the database. Instead, we create a single view log, with a `view count`, and just increment the `view count`.

Screenshot from 2021-05-13 16-02-20

While doing this ticket I noticed an error with updating the view count that was also happening in `UserProfileViewLogs`, so I fixed that - hence so many commits and file changes.

### DIFF
--- a/app/controllers/team_members/journal_entries_controller.rb
+++ b/app/controllers/team_members/journal_entries_controller.rb
@@ -6,7 +6,7 @@ module TeamMembers
     # GET /journal_entries/:id
     def show
       journal_entry
-      view_log
+      log_view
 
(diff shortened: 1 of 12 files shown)


IeuanSkinner (approved):
I've added an additional sort option for the 'Published At' column of this table and I've also added in the subheading stats so no need to do these @AlexW555 👍

PR #95, opened 2021-05-17
Feature - IJ-92 Pyramid of Metrics

IeuanSkinner:
I've created a new partial called "pyramid" that given a wellbeing assessment will be able to generate a pyramid view of this assessment as seen in the below recordings. This view works on mobile too and will display a tooltip with the metric name and the score that user gave for that metric.

Currently this is used in 2 places: TeamMembers > Users > User Profile and TeamMembers > Wellbeing Assessments > Wellbeing Assessment.

In theory we can expose this view to the User > Wellbeing Assessment view too however I've deliberately not done that as it leverages icons as a shorthand for the metrics and it might be that some of these need green lighting with Include UK (specifically the capsules representing addiction).

It's also not wired up to the sliders so can't be used for filling out new Wellbeing Assessments but in theory it can be.

https://user-images.githubusercontent.com/6815945/118506628-64ef2800-b725-11eb-984f-4b4b7bd61cf2.mov

https://user-images.githubusercontent.com/6815945/118506643-6882af00-b725-11eb-86ae-5e4a527d2071.mov


garethdandrews:
Changed the ordering of wellbeing metrics to reflect the pyramid
Screenshot from 2021-05-17 16-55-49

phil-6 (approved):
Love this! Flawless work.

This was designed to be only viewed by team members so I'm happy to leave this as it is for now. Thanks Ieuan!

PR #98, opened 2021-05-20
Feature- IJ-184 User Tags

IeuanSkinner:
I've created a new pair of entities called Tags and UserTags whose relationships to the TeamMember and User entities can be seen below (as well as on LucidChart). The relationship between the TeamMember and Tags entity represents the TeamMember who first created the Tag and the UserTags entity is a joining table between the TeamMember, User and Tag indicating the TeamMember who set the Tag to the User.

IJ User Tagging ERD

Adding this resource creates a new button on the TeamMember dashboard and a new button in the TeamMember navbar.
dashboard

From this button the TeamMember can view an index of all the systems tags along with who created the tag and how many users the Tag is assigned too.
tag_index

By clicking on a Tag the TeamMember can reach the UserTags index view which shows all the Users the Tag is assigned too along with who assigned the Tag and when.
user_tag_index

TeamMembers can view, add and remove Tags from a User from the Users member view.
user_member_view

When the use hits 'Add Tag' a new partially blank Tag appears where the TeamMember can set the Tag from the list of existing Tags.
new_tag_form

dropdown

Which has front-end validation in place.
dropdown_validation

Or the user can write a completely new Tag by clicking on the pencil icon, which also has front-end validation in place.
text_input

text_input_validation


garethdandrews (approved):
Really liking the tags on the user view.

garethdandrews:
I've changed the ordering of user tags from `desc` to `asc` and moved the new tag template to the top of the container. When the user had 5 tags and the add button was clicked (to reveal the template), it wasn't clear if anything had happened as the new tag was out of view.
Screenshot 2021-05-24 at 16 15 34
Also I believe most other resources are sorted by desc by default so keeps things consistent.

garethdandrews:
Do you think we should have a button on this view to add a new tag to be used in the select menu?
Screenshot 2021-05-24 at 16 22 46
If yes I can implement this quickly.

phil-6 (approved):

AlexW555 (approved):

garethdandrews:
I've added the functionality to create tags from the tags index view:
Screenshot from 2021-05-25 15-48-12
Clicking the add button opens a modal:
Screenshot from 2021-05-25 15-58-41
Note that I've also added `titlization` to tags when they're created / compared to clean up the user input:
Screenshot from 2021-05-25 15-59-10

phil-6 (approved):

PR #99, opened 2021-05-21
Feature - IJ-181 Additional User Properties

garethdandrews:
Users can now add additional details to their profile if they chose. 
This work has been based on the UK Governments pattern for Equality Information [https://design-system.service.gov.uk/patterns/equality-information/](url)
The additional fields added: date of birth, pronouns, sex, gender identity, ethnic group, religion and disabilities.
None of the additional fields are required, and are nil by default.
The edit user form became very bloated with the addition of these fields so it has been divided up.
For changing the password, the second password confirmation field is hidden until the user clicks the first password input.

Desktop:
Screenshot from 2021-05-21 12-23-31

Mobile:
Screenshot from 2021-05-21 12-24-07
Screenshot from 2021-05-21 12-24-28

Additionally, since the user needs to input their password to confirm any updates, I have extracted this to a modal that opens when 'Update' is clicked.

Desktop:
Screenshot from 2021-05-21 12-31-13

Mobile: 
Screenshot from 2021-05-21 12-31-25

NOTE: I'm unsure how to correctly store options for select, and for now they are in the html file. I didn't think it was justified adding an entity to the db for things like ethnic group and religion. If anyone can think of a better way let me know and I can change it.


IeuanSkinner:
> NOTE: I'm unsure how to correctly store options for select, and for now they are in the html file. I didn't think it was justified adding an entity to the db for things like ethnic group and religion. If anyone can think of a better way let me know and I can change it.

Typically you would store these options as an entity in the DB so that there's only 1 source for them all. 

If we have pronouns and disabilities be free text fields then we won't be able to effectively use them for classifying people, as in you can't produce analytics for people with 'X' disability as there's no way of linking all the people with 'X' disability (you can do a wildcary query on that free-text field but that could pull in people you don't intend).

If we're not expected to ever produce analytics from these fields however and they're just there to be read and interpreted by the Team Members then they should be fine like this (though I don't think this is the case for disabilities). 

For the pronoun field we could have a mix i.e. 'He/Him, She/Her, They/Them' and then an 'Other' option. Whereas a users disabilities could be one-to-many so I think you probably are going to need a Disabilities entity and a joining table of UserDisabilities where the joining table could either have a lookup to a disability entity or a free text field for the users 'Other' option 🤔

garethdandrews:
> > NOTE: I'm unsure how to correctly store options for select, and for now they are in the html file. I didn't think it was justified adding an entity to the db for things like ethnic group and religion. If anyone can think of a better way let me know and I can change it.
> 
> Typically you would store these options as an entity in the DB so that there's only 1 source for them all.
> 
> If we have pronouns and disabilities be free text fields then we won't be able to effectively use them for classifying people, as in you can't produce analytics for people with 'X' disability as there's no way of linking all the people with 'X' disability (you can do a wildcary query on that free-text field but that could pull in people you don't intend).
> 
> If we're not expected to ever produce analytics from these fields however and they're just there to be read and interpreted by the Team Members then they should be fine like this (though I don't think this is the case for disabilities).
> 
> For the pronoun field we could have a mix i.e. 'He/Him, She/Her, They/Them' and then an 'Other' option. Whereas a users disabilities could be one-to-many so I think you probably are going to need a Disabilities entity and a joining table of UserDisabilities where the joining table could either have a lookup to a disability entity or a free text field for the users 'Other' option thinking

I've updated the pronouns but original ticket says to use free text for disabilities. [https://lilw.atlassian.net/browse/IJ-181?focusedCommentId=10357](url)

phil-6 (approved):
Looks good to me. If Include want any further changes we can do that when they ask for it.

PR #100, opened 2021-05-24
Feature - IJ-174 Guide

AlexW555:
Added in a guide tab for the user dashboard. Currently, it's just screenshots and some basic language text. I have tried to make the language appealing to the end-users, avoiding some technical terms. 

Please can you check through for spelling errors. Or layout tweaks.

https://user-images.githubusercontent.com/17786854/119346521-3972bd00-bc92-11eb-9216-b84d44e23f85.mov

https://user-images.githubusercontent.com/17786854/119346837-9f5f4480-bc92-11eb-8a8f-6d9e69172471.mov


garethdandrews:
It's looking good, I like the style of language and approach you've gone for with this.
I've slightly updated the layout by moving the 'Back To App' button to the top left of the view, and adding margins between the `<hr>` tags and the images:
Screenshot from 2021-05-25 12-05-08

Also did you mean to replace this text?
Screenshot from 2021-05-25 11-54-53

IeuanSkinner:
On mobile the 'Back To App' button is not aligned correctly, I actually think we should get rid of this button as we want users to navigate using the native browser navigation/navbar buttons rather than these in-app back buttons hence why we've removed them in all the other parts of the system.

Screenshot 2021-06-01 at 11 34 04

The guide button is replacing the 'My Voice' button (Include UK might still want this functionality so it's not clear to me that we should be replacing it, a 4th upper-right nav button for the guide that works like the 'Edit Profile'/'Logout' buttons might be better) but the guide itself assumes this button still exists so there's a mismatch occuring here.

AlexW555:
> @AlexW555
> 
> On mobile the 'Back To App' button is not aligned correctly, I actually think we should get rid of this button as we want users to navigate using the native browser navigation/navbar buttons rather than these in-app back buttons hence why we've removed them in all the other parts of the system.
> 
> Screenshot 2021-06-01 at 11 34 04
> 
> The guide button is replacing the 'My Voice' button (Include UK might still want this functionality so it's not clear to me that we should be replacing it, a 4th upper-right nav button for the guide that works like the 'Edit Profile'/'Logout' buttons might be better) but the guide itself assumes this button still exists so there's a mismatch occurring here.

I agree on the button, It takes away from the navbar's purposes and I have removed it.

I also agree with the point on the my voices section, It was only removed when myself and @phil-6 were coding it but I cannot recall the justification at the time. I would say that though that having a guide accessible from these main dashboard buttons will be a better choice than the navbar. 
Screenshot 2021-06-01 at 8 17 42 pm




> @AlexW555
> 
> On mobile the 'Back To App' button is not aligned correctly, I actually think we should get rid of this button as we want users to navigate using the native browser navigation/navbar buttons rather than these in-app back buttons hence why we've removed them in all the other parts of the system.
> 
> Screenshot 2021-06-01 at 11 34 04
> 
> The guide button is replacing the 'My Voice' button (Include UK might still want this functionality so it's not clear to me that we should be replacing it, a 4th upper-right nav button for the guide that works like the 'Edit Profile'/'Logout' buttons might be better) but the guide itself assumes this button still exists so there's a mismatch occuring here.

Agree fully! Changes made

garethdandrews (approved):

IeuanSkinner (approved):
LGTM

PR #102, opened 2021-05-26
Feature - IJ-188 add phone number wellbeing services

phil-6:
Add phone number to the wellbeing services. 
Moved the wellbeing services card to shared/  instead of team members/ as it is used by team members and users.

Also included a fix for a broken link and a small update to the output when seeds run.


garethdandrews (approved):
LGTM

AlexW555 (approved):

PR #104, opened 2021-05-27
Feature- IJ-186 Average for Wellbeing Assessments

garethdandrews:
We now store and display an average score for each wellbeing assessment. The views have been updated in multiple places to display this new information.
Additionally, I have updated the way we display a users wellbeing assessment to make it more consistent with the rest of the application.

#### New Wellbeing Assessment
Before:
Screenshot from 2021-05-27 15-13-28
After:
Screenshot from 2021-05-27 15-18-41

#### Show Wellbeing Assessment
Before:
Screenshot from 2021-05-27 **13-40-31
After:
Screenshot from 2021-05-27 15-20-56

#### Your Services
I decided to move the your services section to the right of the wellbeing assessment as it created a lot of negative space when there were no suggested services. I also added an additional paragraph explaining where to find more services, which also helped to combat the negative space.

Before:
Screenshot from 2021-05-27 15-23-11
After:
Screenshot from 2021-05-27 13-31-57

#### Users Index
Screenshot from 2021-05-27 15-46-18

#### Wellbeing Assessment Index
Screenshot from 2021-05-27 13-34-43

#### User Member Page
Before:
Screenshot from 2021-05-27 13-41-04

After:

With no previous wellbeing assessments:
Screenshot from 2021-05-27 13-28-23

With at least one previous wellbeing assessment, but not for today:
Screenshot from 2021-05-27 15-35-18

Many of the above changes were made to one partial that is used by multiple views. An example on mobile:
Screenshot from 2021-05-27 13-36-37

The sub heading beneath the title containing the metadata (author, created date and created time) is taken from the User Tags section and is a great way of conveying this information. The author of a wellbeing assessment is only shown if it wasn't the user.

#### Wellbeing History Chart
The wellbeing history chart contains an option for average
Screenshot from 2021-06-01 09-11-30

One additional change made was that the wellbeing history chart is now hidden until the user has completed at least 2 wellbeing assessments.

Let me know if you have any thoughts or suggestions!


IeuanSkinner:
The new wellbeing assessment view for users looks like a step back to me, the column for the services is too small so the button text wraps and the yellow text for 'click here' doesn't look good against the light background, I also think having the chart top align with the sliders makes more sense than trying to vertically center align it.

My suggestion for the first problem would be to rethink this view entirely, there's just too much information and controls going on to have this 3 column layout, what if we switched it to a 2 column layout for the chart and controls and then had the services populate in a row above or below these elements?

IeuanSkinner:
@garethdandrews 

I've change the `wellbeing_assessment_history` function so that the last line is the average rather than the first, I've also updated the average pill so that it sits below the main assessment and indicates to the team member that it represents the average.

Screenshot 2021-06-02 at 14 20 04

Screenshot 2021-06-02 at 14 19 49

phil-6:
> @garethdandrews
> 
> The new wellbeing assessment view for users looks like a step back to me, the column for the services is too small so the button text wraps and the yellow text for 'click here' doesn't look good against the light background, I also think having the chart top align with the sliders makes more sense than trying to vertically center align it.
> 
> My suggestion for the first problem would be to rethink this view entirely, there's just too much information and controls going on to have this 3 column layout, what if we switched it to a 2 column layout for the chart and controls and then had the services populate in a row above or below these elements?

I like the idea of having the services appear underneath the wellbeing assessment. We could also use extra space to include the service contact number on the page to give the user more immediate useful information.

IeuanSkinner:
I've updated the wellbeing assessments form so that the services now appear below the chart/controls and the phone numbers are also rendered alongside the link to the website.

https://user-images.githubusercontent.com/6815945/120809121-acf1b580-c541-11eb-9ed8-818ade974c92.mov

phil-6 (approved):

PR #106, opened 2021-06-01
Feature - IJ-172 Sort Functionality For Table Column Headers

IeuanSkinner:
I've updated the sort functionality so that the sort order can be toggled directly from a control on the column header rather than the dropdown which is currently adding bulk to the UI that it needn't. I've also made the sort functionality slightly more generic with a sort param indicating the column on which the user wants to sort and a direction param which we take to be either ascending (asc) or descending (desc).

In this PR I've replaced all the instances of the old sort functionality with this new one but in theory the functionality is now generic enough that we could roll it out to every searchable index, though that would take some time and has not yet been asked for so I think should be a new ticket if needed.

User Profile View Logs Before:
user-profile-view-logs-before

User Profile View Logs After:
user-profile-view-logs-after

Journal Entry View Logs Before:
journal-entry-view-logs-before

Journal Entry View Logs After:
journal-entry-view-logs-after

Tags Index Before:
tags-index-before

Tags Index After:
tags-index-after

User Tags Index Before:
user-tags-index-before

User Tags Index After:
user-tags-index-after


phil-6 (approved):
This looks great to me, thanks Ieuan!

garethdandrews (approved):
LGTM

PR #107, opened 2021-06-02
Feature - IJ-190 Export Wellbeing Assessments

IeuanSkinner:
I've created a new concern called 'Exportable' that can be used to covert any index of information into a CSV/JSON file so long as the model for the resource being exported implements the needed functions, currently this export is only being applied to the wellbeing assessment indexes in the team members scope.

Worth noting that I've also integrated this concern with the functionality from the Pagination concern meaning if the team members wants to manipulate the index with a search query and export the result of this they can! This also means we can update the search controls for the wellbeing assessments index to allow team members to filter based on date range, sex, tags etc. and have the results of that filtering reflected in the files exported from the system so team members can generate datasets like 'WBA of **Women** named **Joanne** in the **Last Month**' and so on. It also means that the additional controls we have for 'Viewed/Not Viewed' and 'Emoji' in the Journal Entries index will be preserve if we end up implementing the exports there too.

The addition of these additional search controls were not in the scope of this ticket however so if we want them we should create a new ticket.

Screenshot 2021-06-02 at 15 14 04

### DIFF
--- a/app/controllers/concerns/exportable.rb
+++ b/app/controllers/concerns/exportable.rb
@@ -1,0 +1,45 @@
+# app/controllers/concerns/exportable.rb
+module Exportable
+  extend ActiveSupport::Concern
+  require 'csv'
+  included do
+    before_action :exportable, only: :export
+  end
+
+  protected
+
+  def exportable
+    @query = exportable_params[:query]
+    @resources = @query.present? ? search : resources
+
+    respond_to do |format|
+      format.csv { send_data to_csv, filename: filename('csv') }
+    end
+  end
+
+  def to_csv
+    CSV.generate(headers: true) do |csv|
+      csv << csv_headers
+
+      @resources.each do |resource|
+        csv << resource.to_csv
+      end
+    end
+  end
+
+  def csv_headers
+    raise 'CSV Headers not overridden'
+  end
+
+  def base_path
+    raise 'Base path not overridden'
+  end
+
+  def filename(type)
+    "#{base_path}-#{DateTime.now.strftime('%Y-%m-%d-%H-%M-%S')}.#{type}"
+  end
+
+  def exportable_params
+    params.permit(:query)
+  end
+end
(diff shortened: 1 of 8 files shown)


IeuanSkinner:
@phil-6 can you let me know if we want the search controls for the wellbeing assessments indexes I outlined in the PR description? If so I can create a ticket and work on integrating them 👍

phil-6:
I think we can leave the search controls for now. I suspect that that might come in useful when we have a more mature analytics brief. 
My reasoning are that at the moment, they can export all and filter themselves when post-processing the data, in the future, when they're looking at a particular data set in-app, then they might want to export it for further analysis.

phil-6 (approved):

IeuanSkinner:
@phil-6 that's an interesting point, as a part of the current exports all I include on the user is their ID and full name so if you wanted to cross-reference for their sex, ethnicity or any tags then the only way to do that is manually checking person by person in the system which isn't very efficient.

Should I add an export function to the users index too so you could cross-reference those 2 outputs? The concern is setup in such a way that this should be pretty simple and it seems like a better solution that bloating the WBA export by including all that additional user info in that export 🤔

phil-6:
My view is that we should include the user data in this export.

It is more intensive on the app at the point of export, but without it I feel as though the data is incomplete and for this style of export I think its a lot more useful to have all the data in one place.

IeuanSkinner:
@phil-6 I've added in the additional fields, I've not put date of birth or release date as though seem like unique categories you can't really group people by (though I suppose you could do some age grouping if you know their date of birth 🤔 ).

Let me know if you think this is now sufficient or needs more info added to the export 👍

phil-6:
I think DoB and Release Date were some of the metrics which Gemma was interested in as it allows her to create "cohorts" based on these.

IeuanSkinner:
@phil-6 these have also been added to the export 👍

phil-6 (approved):

PR #110, opened 2021-06-03
Feature - IJ-182 Minor Updates

IeuanSkinner:
Building off the work @AlexW555 has done I've made some minor style updates that I think add to this work, firstly I've updated the seeds to use 2 `pretty_print` functions for the entity names and the times to ensure the print out is always consistently laid out and not require the developer to trial and error the no. of tab characters needed to get everything to align.

Screenshot 2021-06-03 at 16 25 44

Secondly I've added an additional hidden button to the dashboard so the wellbeing management and tags buttons are the same size as the others, I realise this does leave us with a gap (2 for non-admins) but I still think this looks better (ideally we'd center align these so if you have a single button on the bottom row it looks like the 0 key on a telephone keyboard for example).

Screenshot 2021-06-03 at 16 26 08

Thirdly I've removed showing the value on the wellbeing assessment pill as I was under the impression we wanted to make sure they were describing their state in human rather than mathematical terms (though this might be useful for team members so I'm only hiding it for users).

User:
Screenshot 2021-06-03 at 16 26 55

Team Member:
Screenshot 2021-06-03 at 16 28 03

Lastly I've updated the icons for the wellbeing management button along with the sub-navigation buttons and I've also update the main nav buttons to point to the same locations as the dashboard. I really like the apparent nesting of these 3 indices 👍 

Screenshot 2021-06-03 at 16 28 57

### DIFF
--- a/app/views/shared/_navbar.html.erb
+++ b/app/views/shared/_navbar.html.erb
@@ -15,12 +15,6 @@
       <% unless dashboard_controller? %>
         <% if current_team_member %>
           <div class="navbar-items">
-            <% if current_team_member.admin? %>
-              <%= link_to team_members_path, class: 'btn btn-nav' do %>
-                <i class="fas fa-users"></i>
-                <div class="nav-label hide">Team Members</div>
-              <% end %>
-            <% end %>
             <%= link_to users_path, class: 'btn btn-nav' do %>
               <i class="fas fa-user"></i>
               <div class="nav-label hide">Users</div>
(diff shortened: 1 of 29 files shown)


IeuanSkinner:
@garethdandrews @AlexW555 @phil-6 

As I was working on this I realised that part of the complexity is having to maintain 2 seperate dashboard views in case the team members is an admin or not but would we prefer to keep the dashboards the exact same and instead have the 'Team Members' section linked to via a button in the top-right?

phil-6:
I like the idea of unifying the dashboards. I think a navbar button next to "profile" would  work well for team member management.

I like the rest of the tweaks too. Nice work.

AlexW555 (approved):
Looks all good to me!

IeuanSkinner:
@phil-6 I've now made this change to the dashboard so it's consistent between admin and non-admin team members, the Team Members control in the upper-right is only visible to admins 👍 

Screenshot 2021-06-08 at 09 31 16

garethdandrews (approved):
LGTM

PR #113, opened 2021-06-17
Bugfix - IJ-191 - Reverting changes.

IeuanSkinner:
Reverted the changes made to the description partial as this never took into account the impact on the new form for WBA's. 

I've also made the same changes to the show view as were made for the new view to change how the 'Your Services' buttons appear and look.


phil-6 (approved):
LGTM

PR #116, opened 2021-08-02
Feature - IJ 192 Daily Affirmations

IeuanSkinner:
I've added a new resource type called 'Affirmation' which can be used to schedule a daily affirmation (or 'though for the day' but affirmation is a more succinct way of describing them). I've also updated the seed to be able to generaate upcoming affirmations and archived affirmations (same naming convention as goals), **Note: You will need to update your config.rb file if you want to run the seed to include values for PAST_AFFIRMATIONS and UPCOMING_AFFIRMATIONS, I've gone with 100 and 30 of each.**

The User's will see this affirmation on their dashboard view below the 'Your Journey' header if present.

Screenshot 2021-08-02 at 09 40 03

Team Members have a new button on the dashboard and navbar that takes them to the upcoming affirmations index which will display a table sorted by the scheduled date with today/the next date upcoming at the top.

Screenshot 2021-08-02 at 09 37 47

Screenshot 2021-08-02 at 09 38 05

Screenshot 2021-08-02 at 09 38 22

The archive index contains a reduced table with pagination controls (as this index can get huge whereas you wouldn't expect more than a months worth of upcoming affirmations to be scheduled) which can also be searched according to the affirmation text. (More complex search for the team member and date range could also be added in future if required).

Screenshot 2021-08-02 at 09 38 38


app/assets/stylesheets/fonts.scss
```diff
@@ -1,25 +1,25 @@
 @font-face {
   font-family: 'Montserrat';
-  src: local('Montserrat.ttf');
+  src: font-url('Montserrat.ttf') format('truetype');
```
phil-6, line 3: What's the thought behind this change? I thought we swapped it to `local` to get it working in production?

app/controllers/team_members/affirmations_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+      redirect_back(fallback_location: affirmation_path(@affirmation), flash: error('update'))
+    end
+
+    def create_unique_check
+      return unless Affirmation.where('scheduled_date = ?', affirmation_params[:scheduled_date]).present?
+
+      session[:affirmation_text] = affirmation_params[:text]
+
+      redirect_back(fallback_location: new_affirmation_path, flash: error('create'))
+    end
```
phil-6, line 102: Do we need to do this? 
Instead of having these two checks could we just render a list of the next 30 days, then not allow the date in the form to be user-editable?

I like having the pre-rendered list as it will more easily show when there are gaps.
What do you think about that approach?

app/models/affirmation.rb
```diff
@@ -0,0 +1,13 @@
+# app/models/affirmation.rb
+class Affirmation < ApplicationRecord
+  belongs_to :team_member
+
+  scope :archived, -> { where('scheduled_date < ?', Date.today) }
+  scope :upcoming, -> { where('scheduled_date >= ?', Date.today) }
```
phil-6, line 6: This is lovely!

phil-6 (commented):
Not sure what the best approach for team members to add new affirmations is, but everything else looks good.

app/assets/stylesheets/fonts.scss
IeuanSkinner: This is the fix I did for Dragon Spotter, setting it to local fixed it on production but then when you run it locally the fonts are gone. This method has it working in both contexts.

app/controllers/team_members/affirmations_controller.rb
IeuanSkinner: I hadn't considered that approach, I don't see why that wouldn't work so will change it 👍

IeuanSkinner: @phil-6 I've given it some more thought and not having the date be editable would also impact the archived affirmations as from their a team member can use the edit view if they want to re-use an old affirmation. I could try and work around this by having it be that the date field is only editable if the affirmation is an archived one (i.e. before today) but then these uniqueness checks would still need to exist for that context.

Given this I think it might be better just leave it as is for now, maybe even having an alternate calendar view would be better to see if any days are missing 🤔

phil-6: Could they not just view the archive and the copy and paste the message?
Although that is a bit less convenient, I think the overall UX is probably better from being able to see which upcoming dates are missing.

IeuanSkinner: @phil-6 they could copy+paste but that would lead to duplicate affirmation messages which is slightly redundant and probably not great UX either. Would be able to view the table on a calendar so you easily spot dates that were missing be better? That way nothing needs to functionally change I'll just need to lookup some good calendar UIs.

IeuanSkinner:
@phil-6 can I get your approval on this? I've created a new ticket IJ-199 to capture any enhancement we feel needs to be made to the index view which means the rest of the work should be good as this is the only outstanding discussion point 👍

phil-6 (approved):
Happy with looking at the UX for adding affirmations in a separate ticket.

PR #117, opened 2021-08-03
Feature - IJ 193 Extra User Fields

IeuanSkinner:
Added the extra user fields according to the IJ-193 ticket bar the 'Sentence Length' field as we're still waiting on an update on that. I've removed the _date suffix from all of the date fields as it's redundant and have also set all date fields to date rather than datetime as this just stores redundant data if we don't care about the time.

To access the editable details section of a user a team member can click on the users heading:

Screenshot 2021-08-03 at 10 54 02

They are then presented with a 2 column form with the LHS containing all of the users personal details which are only editable to the user and the RHS containing the additional data that team members can set:

Screenshot 2021-08-03 at 10 53 46

### DIFF
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -230,13 +230,24 @@
   }
 
   &.edit {
-    .col-lg-6 {
-      margin-bottom: 1rem;
+    .col {
       border-radius: 0.5rem;
       padding-bottom: 1rem;
 
       @media screen and (min-width: $screen-md) {
         background: $blue-grad;
+      }
+
+      .col-lg {
+        padding: 0;
+      }
+
+      hr {
+        margin-top: 1rem;
+
+        @media screen and (min-width: $screen-lg) {
+          display: none;
+        }
       }
 
       form {
@@ -263,7 +274,7 @@
         }
 
         input[type="submit"] {
-          margin-top: 0.5rem;
+          margin-top: 1rem;
         }
       }
     }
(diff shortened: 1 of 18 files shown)


IeuanSkinner:
@phil-6 what are you thoughts on removing the '_date' suffix from table columns that are date/datetimes? I've done that in this ticket for users as it just seems redundant however I realise that currently we have this convention on other entities and I'm now questioning whether it's worth it if it's at least consistent 🤔

phil-6:
Comment from stand up - please can we change the edit icon to be a button with some more descriptive words.

phil-6:
> @phil-6 what are you thoughts on removing the '_date' suffix from table columns that are date/datetimes? I've done that in this ticket for users as it just seems redundant however I realise that currently we have this convention on other entities and I'm now questioning whether it's worth it if it's at least consistent thinking

Maybe we should follow the Devise convention, change it to "deleted_at" similar to "created_at" and "updated_at"

IeuanSkinner:
> > @phil-6 what are you thoughts on removing the '_date' suffix from table columns that are date/datetimes? I've done that in this ticket for users as it just seems redundant however I realise that currently we have this convention on other entities and I'm now questioning whether it's worth it if it's at least consistent thinking
> 
> Maybe we should follow the Devise convention, change it to "deleted_at" similar to "created_at" and "updated_at"

That's a great suggestion, I'll make that change across the platform.

IeuanSkinner:
@phil-6 'edit details' link is now below the username and the date fields on the user model now use the _at suffix 👍 

Screenshot 2021-08-04 at 13 00 32

PR #118, opened 2021-08-03
Feature - IJ 196 Update Delete Functionality

IeuanSkinner:
When a user deletes their account rather than removing the user record the record is now marked as deleted and the associated resources which could contain identifying information are removed. Similarly the first and last names are removed along with the email and mobile number for the user.

Demo:

https://user-images.githubusercontent.com/6815945/128047104-d104bca9-fd8c-4aa5-8f90-a9af1d364286.mov

I've also updated the user index cards to show in all-red in the case of a deleted user:

Screenshot 2021-08-03 at 16 51 14

### DIFF
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -2,9 +2,20 @@
 
 .team-members {
   .card {
-    background-color: $grey;
     background: $grey-grad;
     color: $black;
+
+    &.deleted {
+      border-color: $danger;
+
+      .card-header, .card-footer {
+        background: $danger;
+      }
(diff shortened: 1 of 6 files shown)


phil-6:
Wasn't red the same colour as a user with an active crisis event?

IeuanSkinner:
> Wasn't red the same colour as a user with an active crisis event?

Only the header whereas the whole card is red if they're deleted.

Crisis Event:
Screenshot 2021-08-04 at 11 31 56

Deleted:
Screenshot 2021-08-04 at 11 33 01

PR #120, opened 2021-08-16
Feature - IJ-185 Surveys

IeuanSkinner:
This PR encompasses the work I've done to get surveys into the platform and appearing under the 'My Voice' section of the site.

**Seed Changes:** To run the seed you need a new config variable 'SURVEY_RESPONSES' which can be set to either true or false. If false the survey will be created without any programatically generated responses whereas if set to true the responses will be created 1 per user.

This work consists of an update to the user side of this journey to inlcude the submittable survey when they click on 'My Voice' on the dashboard. Note: If the user has only 1 survey available then they will be taken directly to it when they click this button, if they no or many available surveys they will be taken to the surveys index.

Survey as seen on desktop:

Screenshot 2021-08-16 at 09 08 30

Survey as seen on mobile:

Screenshot 2021-08-16 at 09 08 46

When the user views the survey an initial SurveyResponse record is set up indicating a partial response, as the user fills out the survey the corresponding SurveyAnswer and SurveyComment records are created such that the user can leave the survey (even logout altogether) and still come back to it at a later date. Once the user submits the survey the submitted_at  timestamp is set on the survey response and they can no longer edit their response only view it.

Survey index once a survey is completed:

Screenshot 2021-08-16 at 09 11 57

Reviewing that survey show a user their responses without the ability to edit them:

Screenshot 2021-08-16 at 09 12 50

---

The team member portion of the surveys involves being able to create and edit a survey as well as view the responses for a survey. In order to edit a survey you must have no responses so if you have created any you should delete them or re-run the seed.

Surveys dashboard button:

Screenshot 2021-08-16 at 09 15 32

Surveys index:

Screenshot 2021-08-16 at 09 15 46

New survey lets you setup the top-level properties i.e. the name, start and end date for the survey:

Screenshot 2021-08-16 at 09 16 44

The edit survey view allows you to add sections to the survey which in turn can contain questions and comment sections for getting user feedback, the question text and comment section labels will be updated inline so the team member doesn't need to 'save' the form as they're working on it, the same goes for re-ordering the questions, comment-sections and sections themselves.

Screenshot 2021-08-16 at 09 18 39

Note: Re-ordering the questions and comment-sections is easy enough, you simply drag and drop a question onto question or comment-section onto a comment-section and this even works across sections however to re-order a section the whole thing can't be a drag/drop zone as then you get into overlapping drag and drop which is a pain and so instead you simply do it via the header as shown below:

https://user-images.githubusercontent.com/6815945/129533906-4edd223d-e061-47fa-8b88-66af195bcef1.mov

To view all the responses for a survey you simply click on that survey in the index view and you're taken to the member view of the survey which shows an aggregation of all the question responses as bar charts which can be hovered for more detail:

Desktop:

Screenshot 2021-08-16 at 09 29 06

Mobile:

Screenshot 2021-08-16 at 09 29 21

Team members can view an index of all the survey responses by clicking on the View List button and from there they can look at any individuals survey response should they like too:

Screenshot 2021-08-16 at 09 31 45

The comments for a survey have been split out into their own indexes which can be navigated too from the aggregate view:

https://user-images.githubusercontent.com/6815945/129535048-675ba017-3c30-485b-bc96-eb6aea7c1180.mov

Clicking on the survey comment will take the team member to that users entire survey response to view the comment in context.

---

I think the above fully covers the functionality added, there's lots of little ways I think we can improve it but I also think we should maybe leave those for an 'Enhancement' ticket rather than add anymore to the complexity of this already complex piece of work.


phil-6:
Interesting warning from rails:

image

phil-6:
it looks as though when a team member loads the show survey page, we're making an n+1 query, we should be able to improve that I think.

app/views/team_members/surveys/show.html.erb
```diff
@@ -0,0 +1,71 @@
...
+                <div class="col-2 col-lg-1"><div class="header-text"><%= survey_option(index) %></div></div>
+              <% end %>
+            </div>
+
+            <% section.survey_questions.includes(:survey_answers).order(order: :asc).each do |question| %>
+              <div class="row question">
+                <div class="col-12 col-lg-6"><%= question.question %></div>
+                <% total_answers = question.survey_answers.count %>
+                <% 6.times do |index| %>
+                  <% answer_count = question.survey_answers.where(answer: index).count %>
```
phil-6, line 31: This line is giving an n+1 database query which I would very much like to avoid. However this might need a bit of a rethink as to the approach for displaying the completed surveys

phil-6: This might actually be one to add in a TODO comment as its likely to be fine for now, but might cause performance issues in the future

phil-6:
Some interesting unpermitted parameters on the survey comments view.
I've not seen this before but I suspect it is down to how the routes are written.
image

app/views/team_members/survey_responses/show.html.erb
```diff
@@ -0,0 +1,58 @@
...
+            </div>
+
+            <% section.survey_questions.includes(:survey_answers).order(order: :asc).each do |question| %>
+              <div class="row">
+                <div class="col-12 col-lg-6"><%= question.question %></div>
+                <% 6.times do |index| %>
+                  <div class="col-2 col-lg-1">
+                    <label>
+                      <div class="label"><%= survey_option(index) %></div>
+                      <% survey_answer = question.survey_answers.find_by(survey_response: @survey_response) %>
```
phil-6, line 28: Ideally we shouldn't do the `find_by` and variable assignments in the view as it queries the db in every loop. If possible please can we move this to the controller.

app/views/team_members/survey_responses/show.html.erb
```diff
@@ -0,0 +1,58 @@
...
+                  </div>
+                <% end %>
+              </div>
+            <% end %>
+          <% end %>
+
+          <% section.survey_comment_sections.includes(:survey_comments).order(order: :asc).each do |comment_section| %>
+            <div class="row comment-section">
+              <div class="form-floating">
+                <% survey_comment = comment_section.survey_comments.find_by(survey_response: @survey_response) %>
```
phil-6, line 43: As above comment

app/views/team_members/surveys/show.html.erb
phil-6: And maybe add in a "loading" message

phil-6:
When adding new questions, comments or sections to a survey, the page reloads to the top. Is it possible to reload a the same scroll position or page bottom...
Or cheat and add an invisible ID anchor tag and scroll to that?

phil-6:
It would be good to add some validation to ensure that questions and comment sections have validation to ensure that they have headings / questions before a survey goes live

phil-6 (changes requested):
This is a brilliant piece of work. A lot has gone into this and it really shows. Although I've added quite a few comments, they are minor compared to the bulk of what has been created. 
Great work so far Ieuan!

phil-6:
On deleting a survey I get this:
image

phil-6 (approved):
LGTM!

PR #121, opened 2021-08-19
Bugfix - IJ-200 - Crisis Event Mailer Fix

IeuanSkinner:
Fixed the mailer to use root_url rather than root_path so now emails are successfully being generated upon crisis event creation.


IeuanSkinner:
I triggered an email which you should all have receieved, User: Lincoln Collier

AlexW555 (approved):

phil-6 (approved):

PR #123, opened 2021-09-08
Feature - IJ-183 User Streaks/Gamification

IeuanSkinner:
I'm raising this PR now as it's at a nice completion point whereby I _could_ add a lot more to it or this work _could_ exist as is, I've noted at the bottom of this description all the things I think could be added and will put these in an enhancement ticket if we agree.

3 new entities have been setup for this work; Sessions, Achievements and UserAchievements.

IJ Sessions ERD

The Sessions entity is used for tracking the interactions with the system, a unique session is created everyday the user interacts with the system and if the last session created was the day prior to the current session being created we know the user has interacted with the system consecutively and so increment the streak. If this is not the case then a day has passes without interaction and so the streak is reset to 1.

**Note:** I think it might be useful that when the session streak gets reset we go ahead and destroy all the past sessions as these are no longer relevant to the current streak and we don't need them to maintain a record of the total days interacted with the system as their is a cache column on the user record for this. In fact we probably don't need any session records beyond the previous days so we can likely restrict things so at most there's 1/2 session records per user (yesterday + today), thoughts? 

IJ Achievements ERD

The Achievements entity is used to create an achievement which a user can "earn" (be linked to via the UserAchievement entity). Achievements in this case are taken to be counts of some other type of entity namely: sessions, wellbeing assessments, journal entries and achieved goals. 

Each Achievement is tiered into 3 sections bronze, silver and gold, rather than create 3 seperate Achievement records for these tiers and thus duplicating a number of fields across records I've simply taken it as given that every achievement will have these tiers.

Each Achievement can be time-restricted by the Starts At and Ends At columns, this means an achievement could only be achievable in a given month _or_ ir could always be achieveable if these fields are null.

A cached count of all the entities that are associated with achievements are held against the user record, these counts include an all time count and a monthly count. 

Both the generation of the monthly Achievement records **and** the resetting of the monthly counts on user records are handled by scheduled jobs.

All of these achievements are shown in the users new My Journey page as seen in the demo below.

https://user-images.githubusercontent.com/6815945/132516028-5c736dc5-eb63-4944-9c89-78cf92aa8bb8.mov

### Clarification/Discussion Required

- The names and descriptions for each achievement would be useful to solidify, right now I've made them up and used some generic descriptors but Include UK might have better alternatives.
- Should monthly achievements be incremented on a per-record basis or on a per-record per-day basis? By this I mean that currently users can write multiple journal entries in a day and as such can reach the "Journal Journeyman" achievement as quickly as they'd like rather than slowly over the course of the month as they would have to with the "Familiar Face" or "Wellbeing Warrior" achievements.
- Should record deletions decrement the achievement counts? I'm inclined to say yes but this also adds a lot more logic to the back-end as we'd also need to check that if the counts go below the bronze/silver/gold thresholds we remove the achievements that were earned. In this case it's much easier to assume users won't try and abuse the system, especially in the trial phase.

### Future Work

- Trophy Case / Achievements index where users can see all of their previously earned achievements from the months prior. This view can also then be reused for the team members to view a users achievements.


phil-6:
Amazing work Ieuan. This really looks phenomenal!

Will run though the code now but to answer your questions: 

> The names and descriptions for each achievement would be useful to solidify, right now I've made them up and used some generic descriptors but Include UK might have better alternatives.

Are you happy for me to share the gif with them and I'll ask?

> Should monthly achievements be incremented on a per-record basis or on a per-record per-day basis? By this I mean that currently users can write multiple journal entries in a day and as such can reach the "Journal Journeyman" achievement as quickly as they'd like rather than slowly over the course of the month as they would have to with the "Familiar Face" or "Wellbeing Warrior" achievements.

I think there should be a limit to 1 per day for the journals. Maybe we should limit to 5 per day for goals? Although I'm less keen on that. 

> Should record deletions decrement the achievement counts? I'm inclined to say yes but this also adds a lot more logic to the back-end as we'd also need to check that if the counts go below the bronze/silver/gold thresholds we remove the achievements that were earned. In this case it's much easier to assume users won't try and abuse the system, especially in the trial phase.

No. There might be other circumstances which lead to deletion.  I don't think this should influence the achievements. If we're limiting the number of things that can be earned in a day I think that makes it more difficult to game the system.

app/assets/stylesheets/journey.scss
```diff
@@ -0,0 +1,198 @@
...
+          border: 2px solid $brand-primary;
+          color: $brand-primary;
+          padding: 1rem 0;
+
+          .tick {
+            font-size: 1.5rem;
+
+            .tick-flip-panel {
+              background-color: $brand-primary;
+              color: $dark !important;
```
phil-6, line 32: Why do we need the important flag here?

phil-6 (approved):
Amazing work Ieuan. I'm happy for this to get merged as is and further work added as additional tickets

IeuanSkinner:
> Could we simplify it further and just add use current login at field from devise and use .was? to inspect its previous value?

There's already a current_sign_in_at field from Devise which in theory you can get the no. of days the user was signed in for from `(Date.today - User.first.current_sign_in_at.to_date).to_i` however if the user signs in on a different device or just logs out and back in again this would break the streak when it shouldn't.

> What happens if the user has a really long session?

A session record is created for every day when the user interacts with the system so if I interact today and tomorrow then I'll have 2 session records and the streak will be at 2, if on the 3rd day I don't interact I'll have no session record and then when I interact the day after that I will create a session record but now my streak will be reset to 1. 

Keeping these records around allows us to validate this behaviour and we may never *need* to validate this data although we could do _some_ analysis by mapping days when users were active, from that we may see an uptick on weekends or downtick on some particular day 🤷‍♂️ 

> What if a users sessions spans 2 days?

If they interact 2 days in a row they'll have 2 session records, "session" is probably not the best term as what's being mapped here is "interactions" 🤔 

> Do we update the session on login in or last interaction?

Last interaction.

---

> Are you happy for me to share the gif with them and I'll ask?

Sure.

---

> I think there should be a limit to 1 per day for the journals. Maybe we should limit to 5 per day for goals? Although I'm less keen on that.

Can do, we should probably note in the description these behaviours then, something like: "only 1 journal entry per day will count toward this achievement", "upto 5 goals per day can count toward this achievement".

---

> No. There might be other circumstances which lead to deletion. I don't think this should influence the achievements. If we're limiting the number of things that can be earned in a day I think that makes it more difficult to game the system.

Yeah I agree that with the limits in place this becomes less of a problem and there's no real way for it to be perfect, for example a user could still submit a single word journal entry or create and complete a fake goal but there's not a lot we can do about those.

garethdandrews (approved):
LGTM, good work!

PR #132, opened 2021-11-11
Update Form Input Styles

garethdandrews:
Updated the form-floating input styles so error messages are better displayed.
Having some issues getting the checkbox for terms and conditions to highlight when not checked, but without re-writing the field_error_proc there's no easy way to achieve this.
Before:
Screenshot 2021-11-11 at 11 18 24

After:
Screenshot 2021-11-11 at 11 09 38

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -54,6 +54,25 @@
         color: $white;
       }
 
+      .form-control {
+        &:focus {
+          box-shadow: 0 0 0 0.2rem $brand-primary-dark;
+        }
+
+        &.invalid {
+          box-shadow: 0 0 0 0.2rem $danger;
+        }
+      }
+
+      .error-message {
+        position: absolute;
+        top: -0.5rem;
+        right: 0;
+        padding: 1rem 0.75rem;
+        color: $danger;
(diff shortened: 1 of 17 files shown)


app/assets/stylesheets/forms.scss
```diff
@@ -117,6 +140,10 @@
         .form-check-label {
           color: $white;
         }
+
+        .error-message {
```
IeuanSkinner, line 144: Remove this?

IeuanSkinner:
Is there anything simple that can be done to better highlight the 'Terms can't be blank' error? The red on blue text is easy to miss so some form of border highlighting like the text fields have would be useful.

garethdandrews:
> Is there anything simple that can be done to better highlight the 'Terms can't be blank' error? The red on blue text is easy to miss so some form of border highlighting like the text fields have would be useful.

As I said in the description, there wasn't a simple was to highlight it. It would involve a re-write of field_error_proc. 
It is generally common practise on websites these days to omit a checkbox, and instead have the text 'By clicking Sign Up, you agree to our Terms and Conditions.'

IeuanSkinner (approved):
LGTM

PR #147, opened 2022-06-07
Removed reference to current page from param object

FreyaRhianna:
The problem was that send() was lexiographically ordering the pages, and the parameters would always have the page you were on as one parameter (because it was preserving the params in order to also save the feeling filters etc)

So if the latest page you visited started with a higher number, you couldn't then navigate to a page with a lower starting number

One solution may be to override the default behavior, but for now we can just remove the current page from the param object


Zaenilith (approved):
Looks good to me, very simple change that fixes a significant issue

PR #172, opened 2023-01-30
remove released_at instances on ui

oluwatoyosi-swansea:
Remove released_at from ui


AmaraFinbarrs (approved):
Looks good to me!

oluwatoyosi-swansea:
> We also need to remove this data from the database via a migration to remove the column (along with any associations or requirements tied to the release date if there are any)

Okay, will look into this

PR #176, opened 2023-01-30
Enable team members to see the list of contact a specific user has had

oluwatoyosi-swansea:
Enable team members to see the list of contact a specific user has had


oluwatoyosi-swansea:
> Looks really good!
> 
> Generally the main thing I think to point out is that the breadcrumb should reflect the path you took to get there when going from the user perspective e.g., Home > Users > Specific User > Contact Logs rather than Home > Specific User Contact Logs.
> 
> Thinking about usability can we have the page title reflect the user you're looking at in the same way as the breadcrumb currently does when viewing a specific user and can we have the person icon on the card reflect the team member rather than the user (as they will all be the same) when the user id is present in the url, if that makes sense?

Thanks for the comment, i'm looking into this observation and that of the second comment as well

PR #205, opened 2023-03-03
sanitize query parameters

oluwatoyosi-swansea:
sanitize query parameters

### DIFF
--- a/app/controllers/concerns/pagination.rb
+++ b/app/controllers/concerns/pagination.rb
@@ -58,6 +58,7 @@ module Pagination
   end
 
   def wildcard_query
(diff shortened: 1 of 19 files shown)


oluwatoyosi-swansea:
> The sanitization is great but unfortunately it doesn't resolve the brakeman report that we're getting concerning SQL injection: image I've included a screenshot of how the warning appears on our Brakeman report (seen as one of the red markers on the branch checks), this one is marked as particularly vulnerable that we'd like to resolve

Is there a way to run the report again? I'm saying cos i sanitized the `wildcard_query` on the pagination concern so i think that should fix this automatically. Or is there another approach to this particular one?

PR #208, opened 2023-03-15
modify text_area regex to allow newline

oluwatoyosi-swansea:
modify text_area regex to allow newline


Ricky656 (approved):
Simple change, approved!

PR #217, opened 2023-08-04
IJ-294: Implement Upload Feature

AmaraFinbarrs:
This includes an upload feature for both the team members and the user side. It includes the following:

User side
- Users can create, update, view, delete and download an upload.
- Users cannot update an upload that has been approved by a team member.
- Users can search for an upload by its name.

Team Member side
- Team members can create an upload on a specific user.
- Team members can update, view, delete, and download their created uploads.
- Team members can update, view, approve or reject, and download an upload created by a user.
- Any rejected upload automatically gets deleted.
- Team members can search for an upload based on its name and other filter parameters.
- There is an upload activity log feature that records every activity a team member performs on a user's  upload such as viewing, updating, deleting, downloading, rejecting, etc
- Admins can search through upload activity logs using different filter parameters

This fulfils this ticket: [Ticket link](https://lilw.atlassian.net/browse/IJ-294)


Ricky656 (commented):
When I added an upload to a user as an admin, it notes the status of the upload as 'Approved' but in the details section it says "Not yet approved" in the 'approved by' and 'approved at' fields

Ricky656 (commented):
After approving an upload added by a user, as an admin, I can update that upload to hide it from the user - even though they put it up.

Also as an admin I can delete my own uploads on a user, but once I've approved their's I cannot delete it anymore - is this intentional?

These are necessarily bugs, depends what the client wanted.

Ricky656 (commented):
This depends on the client and what they think tbh, but we may need to rename the 'Uploads'  idea, as it's not entirely clear what the purpose of this section even is. 

As someone who wasn't part of the discussion to add this stuff, I don't really know what users or admins are expected to be adding to it - perhaps it should be renamed to just 'Files'? I.E. 'My Files', 'Upload new file'.

Presumably it'll be best to speak to the client about this.

AmaraFinbarrs:
> Also as an admin I can delete my own uploads on a user, but once I've approved their's I cannot delete it anymore - is this intentional?
> 

Yes this is intentional.

PR #221, opened 2023-08-11
[IJ-194] - impelment notifications for team members

oluwatoyosi-swansea:
impelment notifications for team members. Fulfills this ticket: https://lilw.atlassian.net/browse/IJ-194


AmaraFinbarrs (changes requested):
This looks very good and does what it is supposed to!

But it only creates a notification for 6 months after the user signs up. It should also send the notification every 6 months after that. It should notify a team member to check up on the user every 6 months starting from when the user signs up.

Let me know if it already does that so that I can look over the code again. I might have overlooked something.

AmaraFinbarrs (approved):
Looks good, approved!

PR #222, opened 2023-08-11
IJ-318: set allowable upload size

AmaraFinbarrs:
- Allot 5MB for each individual file.
- Set a cap of 250MB for the total file storage per person.
- Let people see how much space they have used.
- Creates a notification for users, when team members add files that they can see.
- Fix bugs and refactor code.

Fufils this ticket: https://lilw.atlassian.net/browse/IJ-318

### DIFF
--- a/app/controllers/team_members/notifications_controller.rb
+++ b/app/controllers/team_members/notifications_controller.rb
@@ -4,7 +4,7 @@ module TeamMembers
     before_action :set_breadcrumbs
 
     def index
-      @notifications = current_team_member.notifications.where(viewed: false)
+      @notifications = current_team_member.notifications.where(viewed: false, upload_id: nil)
     end
 
     def update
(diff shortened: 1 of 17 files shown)


Ricky656 (changes requested):
Looks good so far! I think it's probably worth investigating if we can set the max file size and total file storage as environment variables? It seems like the kind of thing that may be changed based on deployment rather than in the code, let me know what you think. 

Also, do we think it's worth displaying how much storage space is being used up on the main show screen for a user's files?

AmaraFinbarrs:
> Looks good so far! I think it's probably worth investigating if we can set the max file size and total file storage as environment variables? It seems like the kind of thing that may be changed based on deployment rather than in the code, let me know what you think.
> 
> Also, do we think it's worth displaying how much storage space is being used up on the main show screen for a user's files?

Your suggestions make sense. I will implement them.

Ricky656 (changes requested):
I see you've changed the max file size to be set by an environment variable, but the errors and messages that says how much max space you have are hard-coded, so if someone were to change that size with the env variable, it would no longer be accurate.

All the messages that have the hard-coded '5MB' etc, should derive the value from the variable as well

Ricky656 (approved):
Looks good!

PR #224, opened 2023-09-13
IJ-208: Restrict team member view to assigned users

AmaraFinbarrs:
This restricts team members from viewing users that are not assigned to them.

Fulfils this ticket: https://lilw.atlassian.net/browse/IJ-208

### DIFF
--- a/app/controllers/team_members/approvals_controller.rb
+++ b/app/controllers/team_members/approvals_controller.rb
@@ -6,12 +6,6 @@ module TeamMembers
     include Pagination
 
     def index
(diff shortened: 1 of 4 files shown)


oluwatoyosi-swansea (changes requested):
Can you remove the "puts "ANSWER!!! and puts current_team_member.admin"" in the approval controller?
Does the before_action: redirect_back_unless_admin not do the same thing as the validation in the index method?

Apart from these all looks good

oluwatoyosi-swansea (approved):

PR #228, opened 2023-10-12
Feature/ij 308 file and folder manager

oluwatoyosi-swansea:
Solves for this ticket: https://lilw.atlassian.net/browse/IJ-308


Ricky656 (changes requested):
When creating a new file as an admin, the file is successfully created, but I get a 'too many redirects' error 
Screenshot 2023-11-17 at 15 46 10

There is also a 'test folder' that I could save files to by default, but the folder doesn't show up in the folder directory - is it a leftover test folder that needs to be removed?

Ricky656:
Also, folders can only be entered by clicking on the 'open' action, I think clicking on the folder name should also work as most people are going to try to do that

oluwatoyosi-swansea:
> When creating a new file as an admin, the file is successfully created, but I get a 'too many redirects' error Screenshot 2023-11-17 at 15 46 10
> 
> There is also a 'test folder' that I could save files to by default, but the folder doesn't show up in the folder directory - is it a leftover test folder that needs to be removed?

@AmaraFinbarrs please can you help check aswell, can't reproduce this tho

AmaraFinbarrs (approved):
LGTM!

PR #232, opened 2023-11-01
IJ-325: Add OT Score to User Profile Page

AmaraFinbarrs:
This ticket marks a change in how the Occupational Therapist Scores are managed, enhancing the user experience. The display of OT Scores has been moved from the user edit form page to the user show page, where a new card has been introduced specifically for viewing and creating these scores.

[Ticket: IJ-325](https://lilw.atlassian.net/browse/IJ-325)


Ricky656 (changes requested):
I'm not entirely sure on the purpose of this functionality, which makes it a little tough for me to judge, but there are some changes I can suggest:

When adding a new OT score, it gives the 'No.' field as the last highest number, I.E. if there are scores 1, and 2 above it, a new field will also show as '2' until it is saved. Probably need to change that. 

If I go into inspect and remove some of the limitations to the OT score input fields, I can write letters into the field and save the record. The letters aren't kept, which is good, but a new record IS saved using the last numbers - we should make sure it doesn't save any record at all when invalid data like that is inputted.

oluwatoyosi-swansea (approved):
LGTM

Ricky656 (commented):
Most of the code looks good, although make sure you remove the commented out sections of code before actually merging it (user.rb model).

However I can't actually find how to interact with the feature which makes it difficult to judge, only button I can find is the 'Occupational Therapist Assessments' button on the team member dash, but as the db doesn't have any by default it just gives a 'no results' found message and does nothing.

Are these assessments supposed to be created elsewhere, and I'm being dumb, or is this a bug?

Ricky656:
Also, be sure to merge the main branch into this one, and THEN merge the branch into main - to resolve the conflicts without risking messing up main

AmaraFinbarrs:
> Most of the code looks good, although make sure you remove the commented out sections of code before actually merging it (user.rb model).
> 
> However I can't actually find how to interact with the feature which makes it difficult to judge, only button I can find is the 'Occupational Therapist Assessments' button on the team member dash, but as the db doesn't have any by default it just gives a 'no results' found message and does nothing.
> 
> Are these assessments supposed to be created elsewhere, and I'm being dumb, or is this a bug?

Yes. They are created under a specific user, so you can find the feature in a user's profile page though still on the team member side of the app. Also, the feature works specifically when the ENV[ORGANISATION_NAME] is set to 'wallich-journey'. That said, I would need to hide the button you saw on the team member dashboard.

AmaraFinbarrs:
> Also, be sure to merge the main branch into this one, and THEN merge the branch into main - to resolve the conflicts without risking messing up main

Okay, thank you for the heads up. I will be sure to do that.

Ricky656 (dismissed):
When submitting an OT assessment with an incomplete form, it reloads the form with all the selections reset. It'd be much better if it remembered what had been selected already so a user wouldn't have to redo the whole thing when they miss a field; if possible?

Main thing is I hit an error when using the search functionality for the list of OT assessments: 
Screenshot 2023-12-22 at 12 31 16

awhitjon (approved):
Per review session, all looks good.

PR #235, opened 2023-11-03
IJ-203: Custom Error Page Handling

AmaraFinbarrs:
Implemented a custom error page specifically for the 500 internal server error. This choice was made based on the app's configuration, as it is the most likely error the application will encounter.

[Ticket: IJ-203](https://lilw.atlassian.net/browse/IJ-203)


Ricky656 (commented):
Would it be better to use a shared email box for the address rather than using yours specifically? Just thinking about if you're away or anything

oluwatoyosi-swansea (approved):

